PR #6, opened 2019-04-23
Add test query in README.md

BjoernKrollGba:


README.md
```diff
@@ -15,7 +15,7 @@ The project is currently under active development. Please report any issues you
 
 ## Usage
 
-You need Docker to run the LIFE FHIR Store. The most convenient way is to check out this repository and use the `docker-compose.yml` to bring the LIFE FHIR Store up. The default memory requirements are 4 GB. You also need to have port 8080 free on your host.
```
alexanderkiel: I would like to keep default here, because you can change the memory settings and run it with mich less.

README.md
```diff
@@ -76,6 +76,35 @@ If you like to upload your own resources, it's important, that the LIFE FHIR Sto
 
 The most convenient way is to use the [CQL Runner][2]. You have to go into the `Config` menu and set the `CQL Engine` to `http://localhost:8080/cql/evaluate`. The other config options doesn't matter because the CQL Engine of the LIFE FHIR Store always uses its own internal data.
 
+As a test query you can use
+```bash
```
alexanderkiel: We shouldn't use bash here. Please just remove it.

README.md
```diff
@@ -76,6 +76,35 @@ If you like to upload your own resources, it's important, that the LIFE FHIR Sto
...
+library Retrieve1 version '1.0'
+using FHIR version '3.0.0'
```
alexanderkiel: Please use 4.0.0 here. It's now possible.

README.md
```diff
@@ -76,6 +76,35 @@ If you like to upload your own resources, it's important, that the LIFE FHIR Sto
 
 The most convenient way is to use the [CQL Runner][2]. You have to go into the `Config` menu and set the `CQL Engine` to `http://localhost:8080/cql/evaluate`. The other config options doesn't matter because the CQL Engine of the LIFE FHIR Store always uses its own internal data.
 
+As a test query you can use
+```bash
+library Retrieve1 version '1.0'
```
alexanderkiel: One can skip the library line if one doesn't need to import another library.

README.md
```diff
@@ -76,6 +76,35 @@ If you like to upload your own resources, it's important, that the LIFE FHIR Sto
...
+using FHIR version '3.0.0'
+context Population
+
+define NumberOfPatients:
+  Count([Patient])
+
+define Patient:
```
alexanderkiel: It's not a good idea to use `Patient` as name of a definition, because CQL-to-ELM already creates one. Better would be `AllPatients`.

README.md
BjoernKrollGba: No, you cannot run it with less. We have tried to run it with 2 GB and it did not work at all.

alexanderkiel: But you can lower the memory settings of Datomic and the store before running Docker Compose.

BjoernKrollGba: But then we should document how to change the memory settings, properly. Otherwise, it is fairly easy to get the store up without being able to import anything.

alexanderkiel: What about: "The default minimum memory requirements are 4 GB. You have to change the memory settings if you have less memory available."? 

Later we can describe how to change them.

alexanderkiel: And it really works with 2 GB Docker VM. I have tested it.

BjoernKrollGba: It did not work on my machine. Do you remember? I set the available memory of Docker to 2 GB and did not change anything for datomic.

BjoernKrollGba: What about: "The default minimum memory requirements are 4 GB. You have to change the memory settings for Docker and Datomic if you have less memory available."?

alexanderkiel: ...settings for the LIFE FHIR Store and Datomic...

PR #48, opened 2019-10-02
Authentication

benwiz:
## Summary

To add authentication to Blaze, I created a middleware that will try to verify a Signed JWT when the environment variable `OPENID_PROVIDER_URL` is a non-empty string. If `OPENID_PROVIDER_URL` is nil or an empty string, the request will be handled normally.

This PR depends on an external OAuth2 provider to provide and sign the tokens. For example, [Keycloak](https://www.keycloak.org/).

## New Dependencies

- `[buddy/buddy-core "1.6.0"]` for preparing the public key used to verify the token
- `[buddy/buddy-auth "2.2.0"]` for the provided ring middleware

## Authentication functionality

To authenticate the clinical end-user (often, but not always, the patient), [FHIR specifies](http://www.hl7.org/fhir/smart-app-launch/scopes-and-launch-context/index.html#scopes-for-requesting-identity-data) that [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html) is used.

- The url in the `OPENID_PROVIDER_URL` must be the  [OpenID Issuer url](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) to which `/.well-known/openid-configuration` will be appended
- Following the `jwks_uri` property leads to a json wrapped set of keys
- At present, my implementation only looks at the first of these keys
- This key is stored in a Clojure atom and is updated every 60 minutes (currently hardcoded in the Integrant configs)
- The public key from the atom is used to verify the signature of the authorization header bearer token.

### Summary of Changes

I modified `blaze.handler.app/handler` and `blaze.handler.app/router` to accept a collection `middleware` alongside the `handler` collection. This was necessary to pass in configuration data from Integrant.

I expect you will have some opinions about this decision.

### Details

- Specs and configs are added for `authorization` its child `url` in _system.clj_
- Most notably, a `:middleware` property has been added to the `:app-handler` map in [system.clj:198](https://github.com/Breezeemr/blaze/pull/1/files#diff-65f5856be49a96dfce50af7ac7522e33R198-R199)
  - This change is reflected in the `:app-handler` init function at [src/blaze/system.clj:394](https://github.com/Breezeemr/blaze/pull/1/files#diff-65f5856be49a96dfce50af7ac7522e33R394-R395)
  - This same change continues into [src/blaze/handler/app.clj](https://github.com/Breezeemr/blaze/pull/1/files#diff-637d480758abc997cb2ca7597d0caa6b) where all the changes are related to passing the list of middleware into the router. Previous middleware was not configured via Integrant so did not need to be passed in this way, I have not touched any related to other middleware.
  - Lastly, the test in [test/blaze/handler/app_test.clj:37](https://github.com/Breezeemr/blaze/compare/master...Breezeemr:auth?diff=split&expand=1#diff-600823e4ff2df4a145761a133ae4021aR37) must also be updated to account for this change

- Separately, there is an executor running in its own thread that uses `manifold.time/every` to update the OpenID Provider's public key at the given interval

## Questions

- Should other middleware (e.g. `blaze.middleware.json`) that does not depend on external configuration be migrated to the Integrant-oriented approach or left as is?
- Do you have any preferred naming conventions?
- Do you have any commit conventions? Squashing?

### DIFF
--- a/src/blaze/handler/fhir/core.clj
+++ b/src/blaze/handler/fhir/core.clj
@@ -7,15 +7,18 @@
     [ring.util.response :as ring]))
 
 
-(defn router [base-url conn handlers]
+(defn router [base-url conn handlers middleware]
   (reitit-ring/router
     ["" {:blaze/base-url base-url}
-     ["" {:post (:handler.fhir/transaction handlers)}]
+     [""
+      {:middleware [(:middleware/guard middleware)]
+       :post (:handler.fhir/transaction handlers)}]
      ["metadata"
       {:get (:handler.fhir/capabilities handlers)}]
-     ["_history"
-      {:get (:handler.fhir/history-system handlers)}]
-     ["{type}" {:middleware [[wrap-type conn]]}
+    ["_history"
+      {:middleware [(:middleware/guard middleware)]
+       :get (:handler.fhir/history-system handlers)}]
+     ["{type}" {:middleware [(:middleware/guard middleware) [wrap-type conn]]}
       [""
        {:name :fhir/type
         :get (:handler.fhir/search handlers)
@@ -35,7 +38,8 @@
          {:name :fhir/versioned-instance
           :get (:handler.fhir/read handlers)}]]]]
      ["Measure/{id}/$evaluate-measure"
-      {:get
+      {:middleware [(:middleware/guard middleware)]
+       :get
        (:handler.fhir.operation/evaluate-measure handlers)
        :post
        (:handler.fhir.operation/evaluate-measure handlers)}]]
@@ -88,16 +92,21 @@
                 :handler.fhir.operation/evaluate-measure]))
 
 
+(s/def ::middleware
+  (s/keys :req [:middleware/guard]))
+
+
 (s/def :handler.fhir/core fn?)
 
 
 (s/fdef handler
-  :args (s/cat :base-url string? :conn ::ds/conn :handlers ::handlers)
+  :args (s/cat :base-url string? :conn ::ds/conn :handlers ::handlers :middleware ::middleware)
   :ret :handler.fhir/core)
+
 
 (defn handler
   "Whole app Ring handler."
-  [base-url conn handlers]
+  [base-url conn handlers middleware]
   (reitit-ring/ring-handler
-    (router base-url conn handlers)
+    (router base-url conn handlers middleware)
     default-handler))
--- a/src/blaze/middleware/authentication.clj
+++ b/src/blaze/middleware/authentication.clj
@@ -1,0 +1,58 @@
+(ns blaze.middleware.authentication
+  "Verifies a signed JWT using OpenID Connect to provide
+   the public key used to sign the token."
+  (:require
+    [buddy.auth.backends :as backends]
+    [buddy.auth.middleware :as middleware]
+    [buddy.core.keys :as keys]
+    [cheshire.core :as json]
+    [clojure.spec.alpha :as s])
+  (:import
+    (java.security PublicKey)))
+
+
+(s/fdef public-key
+  :args (s/cat :jwks-json string?)
+  :ret (s/nilable #(instance? PublicKey %)))
+
+
+(defn public-key [jwks-json]
+  "Take a the first jwk from jwks-json string and
+   convert it into a PublicKey."
+  (some-> jwks-json
+          (json/parse-string keyword)
+          :keys
+          first
+          keys/jwk->public-key))
+
+
+(s/fdef jwks-json
+  :args (s/cat :url string?)
+  :ret map?)
+
+
+(defn jwks-json [url]
+  (let [well-known "/.well-known/openid-configuration"
+        jwks-json  (some-> url
+                           (str well-known)
+                           slurp
+                           json/parse-string
+                           (get "jwks_uri")
+                           slurp)]
+    (if (some? jwks-json)
+      jwks-json
+       (throw (ex-info "No jwk found"
+                       {:cause (str "No jwk found at " url well-known)})))))
+
+
+(s/fdef wrap-authentication
+  :args (s/cat :public-key #(instance? PublicKey %)))
+
+
+(defn wrap-authentication [public-key]
+  (fn [handler]
+    (middleware/wrap-authentication
+     handler
+     (backends/jws {:token-name "Bearer"
+                    :secret     public-key
+                    :options    {:alg :rs256}}))))
--- a/src/blaze/middleware/guard.clj
+++ b/src/blaze/middleware/guard.clj
@@ -1,0 +1,14 @@
+(ns blaze.middleware.guard
+  (:require
+   [buddy.auth :refer [authenticated?]]
+   [ring.util.response :as ring]))
+
+
+(defn wrap-guard
+  "If the request is unauthenticated return a 401 response."
+  [handler]
+  (fn [request]
+    (if (authenticated? request)
+      (handler request)
+      (-> (ring/response {:message "Unauthenticated"})
+          (ring/status 401)))))
--- a/src/blaze/system.clj
+++ b/src/blaze/system.clj
@@ -33,6 +33,8 @@
     [blaze.metrics :as metrics]
     [blaze.middleware.fhir.metrics :as fhir-metrics]
     [blaze.middleware.json :as json]
+    [blaze.middleware.authentication :as authentication]
+    [blaze.middleware.guard :as guard]
     [blaze.server :as server]
     [blaze.structure-definition :refer [read-structure-definitions]]
     [blaze.terminology-service.extern :as ts]
@@ -57,6 +59,8 @@
 (s/def :config/logging (s/keys :opt [:log/level]))
 (s/def :database/uri string?)
 (s/def :config/database-conn (s/keys :opt [:database/uri]))
+(s/def :openid-provider/url string?)
+(s/def :config/authentication (s/keys :opt [:openid-provider/url]))
 (s/def :term-service/uri string?)
 (s/def :term-service/proxy-host string?)
 (s/def :term-service/proxy-port pos-int?)
@@ -85,6 +89,7 @@
     :opt-un
     [:config/logging
      :config/database-conn
+     :config/authentication
      :config/term-service
      :config/cache
      :config/fhir-capabilities-handler
@@ -106,6 +111,11 @@
    :database-conn
    {:structure-definitions (ig/ref :structure-definitions)
     :database/uri "datomic:mem://dev"}
+
+   :authentication {}
+
+   :guard
+   {:authentication (ig/ref :authentication)}
 
    :term-service
    {:uri "http://tx.fhir.org/r4"}
@@ -179,13 +189,17 @@
      :handler.fhir/transaction (ig/ref :fhir-transaction-handler)
      :handler.fhir/update (ig/ref :fhir-update-handler)
      :handler.fhir.operation/evaluate-measure
-     (ig/ref :fhir-operation-evaluate-measure-handler)}}
+     (ig/ref :fhir-operation-evaluate-measure-handler)}
+    :middleware
+    {:middleware/guard (ig/ref :guard)}}
 
    :app-handler
    {:handlers
     {:handler/cql-evaluation (ig/ref :cql-evaluation-handler)
      :handler/health (ig/ref :health-handler)
-     :handler.fhir/core (ig/ref :fhir-core-handler)}}
+     :handler.fhir/core (ig/ref :fhir-core-handler)}
+    :middleware
+    {:middleware/authentication (ig/ref :authentication)}}
 
    :server-executor {}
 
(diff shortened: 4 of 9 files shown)


alexanderkiel:
First thanks for your very good documented PR.

To answer your questions:

>Should other middleware (e.g. blaze.middleware.json) that does not depend on external configuration be migrated to the Integrant-oriented approach or left as is?

I would left them like they are because they don't require configuration or any other lifecycle.

>Do you have any preferred naming conventions?

You are quite align with what I would consider naming conventions, except `[clojure.string :as string]` should be `[clojure.string :as str]`.

>Do you have any commit conventions? Squashing?

Yes I like to have all commit squashed, except if they improve understanding of the PR a lot. Here the PR is small enough for only one commit. Also please note that I use git-flow in Blaze. So please rebase on develop.

Some questions and recommendations from my side:

* Why don't you use `buddy-auth`? It already brings `buddy.auth.middleware/wrap-authorization` middleware. So you safe a lot of code.
* I use [manifold](https://github.com/ztellman/manifold) and [aleph](https://github.com/ztellman/aleph) for async I/O. You do I/O to fetch the public key. You can't do this blocking in the middle of a request/response cycle. With an aleph HTTP server, you can return a manifold deferred instead of a concrete response. However, I even wouldn't fetch the public key during a normal request/response cycle at all because if the auth provider is slow, that would make Blaze responses slow. Even if we do that only once every hour, it's still a candidate for bad long tail response times. I would take an executor service with a single thread and spawn the key refresh into it.
* I would like to see tests of the atom update function.
* I would rename the config param :authorization/url into :openid-provider/url and append the path `/.well-known/openid-configuration` to it.

I like the fact that authentication is optional and how you integrated it into Blaze. So let's get my main points out of the way and move forward.

benwiz:
Thanks for the feedback and suggestions, Alex. I'll let you know when I'm ready for another review

benwiz:
I've made the requested changes:

- Append `/.well-known/openid-configuration` to the issuer url ([authentication.clj:22](https://github.com/life-research/blaze/pull/48/files#diff-d529b9e06fcc59bec6a63a8079c5cf89R22))
- Use buddy-auth's provided middleware ([authentication.clj:70](https://github.com/life-research/blaze/pull/48/files#diff-d529b9e06fcc59bec6a63a8079c5cf89R70))
- Update the public key atom in its own thread outside of the request flow ([authorization.clj:48](https://github.com/life-research/blaze/pull/48/files#diff-d529b9e06fcc59bec6a63a8079c5cf89R48), [system.clj:270](https://github.com/life-research/blaze/pull/48/files#diff-65f5856be49a96dfce50af7ac7522e33R270))
- Create a test to check the public key updates ([openid_provider_test.clj](https://github.com/life-research/blaze/pull/48/files#diff-65f5856be49a96dfce50af7ac7522e33R270))
  - openidconnect.net provides an example OpenID configuration which I have chosen to use here for the test

@alexanderkiel enjoy your weekend

benwiz:
Associated PR for docs https://github.com/life-research/blaze-doc/pull/2

alexanderkiel:
Thank you. I didn't manage to look at your PR today. Will do it tomorrow.

alexanderkiel:
# Public Key Rotation

I must admit, that we don't use public key rotation, so I did not see that buddy auth doesn't support it very well.

The `wrap-authentication` function is only called once at startup not once per request. So you dereference the `public-key-atom` only once. All updates to the atom are not seen by the JWS backend.

The only solution I see is to reimplement the JWS backend to be able to dereference the atom at every request.

Aside from that, I don't think that public key rotation with only one key will work at all. Let me explain why: If you only have one key and you do not poll the provider synchronously, which isn't possible, you will face the situation were the provider already signs with a new key but you still have the old one. That will result in denying valid JWT's. Instead, you will need at least two keys, one currently primary key and one secondary one. When the provider rotates the keys, it will nominate the secondary key to the new primary key and generate a new secondary key. Blaze will always use both keys to do the validation. After the rotation, the secondary key will validate. So no validation gap will exist. If I'm correct, we will need to save all keys provided in the openid configuration and use all for validation.

# Rotation Itself Using mt/every

From the doctoring of `mt/every`:

>If the invocation of `f` ever throws an exception, repeated invocation is automatically cancelled.

We have to handle exceptions during the fetch of the public key. Otherwise `mt/every` will stop executing our function. I would just catch all exceptions inside `public-key-atom-value` and log a warning.

To be able to monitor whether the polling is still running, we should use [prom-metrics](https://github.com/alexanderkiel/prom-metrics). I don't know if you use Prometheus for metrics, but Blaze already provides various metrics.

Additionally `mt/every` returns a function which can be called to cancel the execution. We should call this function at Integrant shutdown time using `ig/halt-key!` like with the HTTP servers. You can find all that in the `blaze.system` namespace. Shutting down such things is especially important in development were you will restart the system several times.

# Check of Successful Authentication

Currently you check for successful authentication right inside the middleware. I would suspect that at least the CapabilityStatement under `[base]/metadata` should be accessible unauthenticated. So I would split both middleware and put the checking middleware only at the endpoints needed. 

# Tests

The test namespace should have the same name as the productive namespace adding `-test`. I also would like to see unit tests instead of this integration test. I don't think that we will benefit much from integration testing the fetching of a JSON document from a remote URL. It would be sufficient to test that we successful extract a `BCRSAPublicKey` from a captured JSON. I would also not test the atom swapping. A `Thread/sleep` is already a code smell in tests.

benwiz:
Regarding public key rotation...

Since it is not requirement for you and not necessary for us I can remove the feature and load the public key when the server starts. If there is an error, the system will exit to avoid exposing any data. Do you have a preferred way method for exiting the system? If not, I'll use `System/exit`

alexanderkiel:
If you throw an exception during Integrant initialization, (System/exit 1) is called automatically.
https://github.com/life-research/blaze/blob/db1e0131209fdfd0975722dc63dcd5bb290f69ee/src/blaze/core.clj#L31
So you can just slurp the public key inside `ig/init-key` and exceptions are handled that way.

alexanderkiel:
Sorry for closing this. I just used the wrong button.

benwiz:
Hi @alexanderkiel 

I've made the following changes:

- [guard.clj](https://github.com/life-research/blaze/pull/48/files#diff-9dd651949a7b69663e647740d694f689) is a new middleware that will reject an unauthorized request if the request has passed through the authorization middleware (if no `OPENID_PROVIDER_URL` env var is set, `wrap-guard` will have no effect)
- [src/blaze/handler/fhir/core.clj](https://github.com/life-research/blaze/pull/48/files#diff-da0cb36f8defa7dc16de10554c6c18c0) uses the new `wrap-guard` middleware in several but not all routes
- [authentication_test.clj](https://github.com/life-research/blaze/pull/48/files#diff-50917b0c0be974650f9659ed735b7822) now tests converting a jwks json string to a BCRSAPublicKey object.

src/blaze/middleware/authentication.clj
```diff
@@ -0,0 +1,42 @@
...
+  (:require
+   [buddy.auth.backends :as backends]
+   [buddy.auth.middleware :as middleware]
+   [buddy.core.keys :as keys]
+   [cheshire.core :as json]
+   [manifold.time :as mt]
+   [ring.util.response :as ring]))
+
+
+(defn public-key [jwks-json]
```
alexanderkiel: Would like to see the spec directly here, not only in tests. Every Public function should have a spec:
```clojure
(s/fdef public-key
  :args (s/cat :jwks-json string?)
  :ret (s/nilable #(instance? PublicKey %)))
```
Also a short doc would be good.

src/blaze/middleware/authentication.clj
```diff
@@ -0,0 +1,42 @@
+(ns blaze.middleware.authentication
+  "Ring middleware that verifies a signed JWT using OpenID Connect
+   to provide the public key used to sign the token. The public key
+   is stored in an atom that is updated every `expiration-minutes`."
+  (:require
+   [buddy.auth.backends :as backends]
```
alexanderkiel: I use two-spaces indentation here.

src/blaze/middleware/authentication.clj
```diff
@@ -0,0 +1,42 @@
...
+   [buddy.auth.middleware :as middleware]
+   [buddy.core.keys :as keys]
+   [cheshire.core :as json]
+   [manifold.time :as mt]
```
alexanderkiel: unused import

test/blaze/middleware/authentication_test.clj
```diff
@@ -0,0 +1,34 @@
...
+   [clojure.spec.test.alpha :as st]
+   [clojure.test :refer :all]
+   [taoensso.timbre :as log])
+  (:import
+   (org.bouncycastle.jcajce.provider.asymmetric.rsa BCRSAPublicKey)))
+
+
+(defn fixture [f]
+  (st/instrument)
+  (st/instrument
```
alexanderkiel: not needed because of spec in prod

test/blaze/middleware/authentication_test.clj
```diff
@@ -0,0 +1,34 @@
...
+
+
+(deftest public-key-test
+  (testing "Returns type BCRSAPublicKey after parsing jwks-json string"
+    (let [k (public-key jwks-json)]
+      (is (= BCRSAPublicKey (type k))))))
```
alexanderkiel: I would rather test in the interface `java.security.PublicKey` here

src/blaze/middleware/authentication.clj
```diff
@@ -0,0 +1,42 @@
...
+  (let [well-known "/.well-known/openid-configuration"
+        jwks-json  (some-> url
+                           (str well-known)
+                           slurp
+                           json/parse-string
+                           (get "jwks_uri")
+                           slurp)]
+    (if (some? jwks-json)
+      jwks-json
+      (throw (ex-info (str "No jwk found at " url well-known))))))
```
alexanderkiel: `ex-info` needs at least two args

src/blaze/system.clj
```diff
@@ -253,6 +261,13 @@
   (d/connect uri))
 
 
+(defmethod ig/init-key :authentication
+  [_ {:openid-provider/keys [url]}]
+  (let [jwks-json  (when (not-empty url) (authentication/jwks-json url))
```
alexanderkiel: I would write:

```clojure
(if (str/blank? url)
    identity
    (do (log/info "Enabled authentication using OpenID provider:" url)
        (-> url
            authentication/jwks-json
            authentication/public-key
            authentication/wrap-authentication)))
```

Returning `identity` in case we have no url is the no-op middleware.

src/blaze/middleware/authentication.clj
```diff
@@ -0,0 +1,42 @@
...
+
+
+(defn wrap-authentication [public-key]
```
alexanderkiel: That can be:

```clojure
(s/fdef wrap-authentication
  :args (s/cat :public-key #(instance? PublicKey %)))

(defn wrap-authentication [public-key]
  (fn [handler]
    (middleware/wrap-authentication
      handler
      (backends/jws {:token-name "Bearer"
                     :secret public-key
                     :options {:alg :rs256}}))))
```

src/blaze/middleware/authentication.clj
```diff
@@ -0,0 +1,42 @@
+(ns blaze.middleware.authentication
+  "Ring middleware that verifies a signed JWT using OpenID Connect
```
alexanderkiel: That is outdated.

src/blaze/handler/fhir/core.clj
```diff
@@ -35,7 +39,8 @@
          {:name :fhir/versioned-instance
           :get (:handler.fhir/read handlers)}]]]]
      ["Measure/{id}/$evaluate-measure"
-      {:get
+      {:middlewarep [wrap-guard]
```
alexanderkiel: typo

src/blaze/middleware/guard.clj
```diff
@@ -0,0 +1,16 @@
...
+  (:require
+   [buddy.auth :refer [authenticated?]]
+   [ring.util.response :as ring]))
+
+
+(defn wrap-guard
+  "If the request is unauthenticated return an  response."
+  [handler]
+  (fn [request]
+    (if (:tried-authentication request)
```
alexanderkiel: I would remove the condition and make the whole guard middleware `identity` if we have no authentication enabled.

src/blaze/system.clj
```diff
@@ -107,6 +111,8 @@
    {:structure-definitions (ig/ref :structure-definitions)
     :database/uri "datomic:mem://dev"}
 
+   :authentication {}
```
alexanderkiel: The optional auth-guard middleware would also need a key here.

benwiz:
I made the requested changes. Most significantly Integrant now provides `identity` for `wrap-guard` if the `authentication` key is the `identity` fn. 

For _app_test.clj_ and _core_test.clj_ I'm passing in the `identity` fn as the middleware function because the test is focusing on the route matching rather than the middleware functionality. Is this correct?

Issue #185, opened 2020-07-15
Add Diagnostics Message to Object Not Found Issues


fheuschkel:
Improve Operations Outcome on not found Ressource.
Add Diagnostics Message with FHRIR Path.

alexanderkiel:
Closed due to lag of current interest.

PR #187, opened 2020-07-16
Add Diagnostics Message to Object Not Found Issues #185

fheuschkel:
Improve Operations Outcome on not found Ressource.
Add Diagnostics Message with FHRIR Path.

### DIFF
--- a/modules/interaction/src/blaze/interaction/read.clj
+++ b/modules/interaction/src/blaze/interaction/read.clj
@@ -32,7 +32,7 @@
   (str "W/\"" (-> resource :meta :versionId) "\""))
 
 
-(defn- db [node vid]
+(defn- db [node vid type id]
   (cond
     (and vid (re-matches #"\d+" vid))
     (let [vid (Long/parseLong vid)]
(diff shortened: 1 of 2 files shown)


modules/interaction/src/blaze/interaction/read.clj
```diff
@@ -41,7 +41,8 @@
     vid
     (md/error-deferred
       {::anom/category ::anom/not-found
-       :fhir/issue "not-found"})
+       :fhir/issue "not-found"
+       ::anom/message (format "Version ID `%s` is Invalid." vid)})
```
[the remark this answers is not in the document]
alexanderkiel: The main problem here is, that I don't like to tell the user that the `vid` is invalid. The `vid` is opaque to the user. I would rather only tell him that there is no such version of the resource with this `vid`. So please hand over the `db` function `type` and `id` and report that the resource `<type>/<id> with versionId <vid>` was not found.

modules/interaction/src/blaze/interaction/read.clj
```diff
@@ -41,6 +41,7 @@
     vid
     (md/error-deferred
       {::anom/category ::anom/not-found
+       ::anom/message (format "Resource ´/%s/%s´ with versionId ´%s´ was not found." type id vid)
```
alexanderkiel, line 44: This should be backticks here.

fheuschkel: Done

PR #721, opened 2022-05-18
Implement CQL ConvertsToString

marlenaMyr:


modules/cql/test/blaze/elm/compiler/type_operators_test.clj
```diff
@@ -571,6 +571,78 @@
...
+
+  (testing "Integer"
+    (are [x] (true? (core/-eval (tu/compile-unop elm/converts-to-string elm/integer x)
+                                    {} nil nil))
+      "-1"
+      "0"
+      "1"))
+
+  (testing "Decimal"
+    (are [x] (true? (core/-eval (tu/compile-unop elm/converts-to-string elm/decimal x)
```
alexanderkiel, line 589: I would only use one test value for each type, because the result only depends on the type and not on the actual value.

modules/cql/test/blaze/elm/compiler/type_operators_test.clj
```diff
@@ -571,6 +571,78 @@
...
 ;;
 ;; If the argument is null, the result is null.
+(deftest compile-converts-to-string-test
+  (testing "Boolean"
+    (are [x] (true? (core/-eval (tu/compile-unop elm/converts-to-string elm/boolean x)
+                                    {} nil nil))
+      "true"
+      "false"))
+
+  (testing "Integer"
```
alexanderkiel, line 581: Please add a test for `Long` and `String`.

modules/cql/test/blaze/elm/compiler/type_operators_test.clj
```diff
@@ -571,6 +571,78 @@
...
+      "0.0001"
+      "0.00001"
+      "0.000001"
+      "0.0000001"
+      "0.00000001"
+      "0.000000001"
+      "0.000000005"))
+
+  (testing "Quantity"
+    (are [x] (true? (core/-eval (tu/compile-unop elm/converts-to-string elm/quantity
```
alexanderkiel, line 608: You can remove the `core/-eval` part on the static tests.

alexanderkiel (changes requested):

PR #747, opened 2022-06-20
Implement CQL ToTime and rearrange ToDate and ToDateTime

marlenaMyr:


modules/cql/src/blaze/elm/date_time.clj
```diff
@@ -1360,3 +1345,23 @@
...
+    (.toLocalTime this))
+
+  OffsetDateTime
+  (to-time [this now]
+    (-> (.withOffsetSameInstant this (.getOffset ^OffsetDateTime now))
+        (.toLocalTime)))
+
+  PrecisionLocalTime
+  (to-time [this _]
+   (.-local_time this)))
```
alexanderkiel, line 1367: Every file should end with a newline. Please add it here.

alexanderkiel (changes requested):

PR #790, opened 2022-09-13
Switch to Ubuntu 20.04 to Prevent Clone3 Call

alexanderkiel:
As one can read in this [comment][1] and this [commit][2] the glibc of
Ubuntu 22.04 uses a clone3 call which older versions of Docker do not
fail with the right error number. So on older versions of Docker the JVM
can't create threads.

[1]: <https://github.com/adoptium/containers/issues/215#issuecomment-1142046045>
[2]: <https://github.com/moby/moby/commit/9f6b562dd12ef7b1f9e2f8e6f2ab6477790a6594>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM eclipse-temurin:17-jre
+FROM eclipse-temurin:17-jre-focal
 
 RUN apt-get update && apt-get upgrade -y && \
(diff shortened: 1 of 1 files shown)


alexanderkiel:
I had no requests for this older image. So I'll close this PR.

PR #792, opened 2022-09-14
Use Manual WAL Flash at Index Database

alexanderkiel:


alexanderkiel:
I saw no improvement doing manual WAL flush.

PR #793, opened 2022-09-15
Revert to use Default Number Of RocksDB Threads

alexanderkiel:


alexanderkiel:
I didn't find any improvement by reverting to the default number of threads.

PR #795, opened 2022-09-19
Implement CQL Is

marlenaMyr:


modules/cql/src/blaze/elm/compiler/type_operators.clj
```diff
@@ -228,7 +228,65 @@
     (->DescendentsOperatorExpression source)))
 
 
-;; TODO 22.18. Is
+;; 22.18. Is
+(defrecord IsExpression [operand type pred]
+  core/Expression
+  (-eval [_ context resource scope]
+    (pred (core/-eval operand context resource scope)))
```
alexanderkiel, line 235: I made sure, that we can just call `prep` here without doing anything additionally because of performance.

modules/cql/src/blaze/elm/compiler/type_operators.clj
```diff
@@ -228,7 +228,65 @@
...
+    "Quantity" ['elm/quantity quantity/quantity?]
+    "String" ['elm/string string?]))
+
+
+(defn- matches-named-type-is [type-name]
+  (let [[type-ns type-name] (elm-util/parse-qualified-name type-name)]
+    (case type-ns
+      "http://hl7.org/fhir"
+      [(symbol "fhir" type-name)
+       (let [fhir-type (keyword "fhir" type-name)]
```
alexanderkiel, line 257: I had to change the match function here, because the set returns either the keyword if it matches or nil, but we need boolean here.

modules/cql/src/blaze/elm/compiler/type_operators.clj
```diff
@@ -228,7 +228,65 @@
...
+    (matches-named-type-is is-type)
+
+    is-type-specifier
+    (matches-type-specifier-is is-type-specifier)))
+
+
+(defmethod core/compile* :elm.compiler.type/is
+  [context {:keys [operand] :as expression}]
+  (let [[type pred] (matches-type-is expression)]
+    (->IsExpression (core/compile* context operand) type pred)))
```
alexanderkiel, line 288: I made sure, we create always an `IsExpression` because I like to have the form also on `nil` values and the is-expression should never return `nil`.

marlenaMyr: Thank you!

PR #839, opened 2022-11-15
Implement CQL concept

marlenaMyr:


modules/cql/src/blaze/elm/compiler/clinical_values.clj
```diff
@@ -72,8 +73,14 @@
 
 
 ;; 3.6. Concept
-;;
-;; TODO
+(defn- compile-codes [context codes]
+  (map #(core/compile* context %) codes))
+
+(defmethod core/compile* :elm.compiler.type/concept
+  [context
```
alexanderkiel, line 80: You don't need a newline here.

modules/cql/src/blaze/elm/compiler/clinical_values.clj
```diff
@@ -72,8 +73,14 @@
...
+   {:keys [codes]}]
+  (let [codes_comp (compile-codes context codes)]
```
alexanderkiel, line 82: I would remove the `let` here, because everything fits on one line and the binding name doesn't add much clarity because the function name `compile-codes` says all.

modules/cql/src/blaze/elm/concept.clj
```diff
@@ -0,0 +1,16 @@
+(ns blaze.elm.concept
+    "Implementation of the cocept type."
```
alexanderkiel, line 2: The intendation should be only 2 chars here. There is a typo in "cocept".

modules/cql/src/blaze/elm/concept.clj
```diff
@@ -0,0 +1,16 @@
...
+
+
+(defrecord Concept [codes]
+  p/Equivalent
+  ; todo
+  )
+
+
+(defn to-concept
+  "Returns a CQL concept whitch isn't the same as a FHIR concept from the database."
```
alexanderkiel, line 14: whitch -> which and I would just remove the comparison with FHIR.

modules/cql/src/blaze/elm/spec.clj
```diff
@@ -316,6 +318,20 @@
   :elm/code-system-ref)
 
 
+;; 3.6. Concept
```
alexanderkiel, line 321: very good :-)

modules/cql/test/blaze/elm/compiler/clinical_values_test.clj
```diff
@@ -136,7 +137,65 @@
...
 ;;
 ;; The Concept type represents a literal concept selector.
-;; TODO
+(deftest compile-concept-test
+  (testing "without version and one code"
+    (let [context
+          {:library
+           {:codeSystems
+            {:def [{:name "sys-def-115852" :id "system-115910"}]}}}]
+      (given (c/compile context #elm/concept [[["sys-def-115852" "code-115927"]]])
```
alexanderkiel, line 146: The intendation should be only 2 chars here.

modules/cql/test/blaze/elm/compiler/clinical_values_test.clj
```diff
@@ -136,7 +137,65 @@
...
+          {:library
+           {:codeSystems
+            {:def
+             [{:name "sys-def-120434"
+               :id "system-120411"
+               :version "version-120408"}]}}}]
+      (given (c/compile context #elm/concept [[["sys-def-120434" "code-115927"]]])
+             type := Concept
+             [:codes 0 type] := Code
+             [:codes 0 :system] := "system-120411"
```
alexanderkiel, line 177: The version is missing here.

modules/cql/test/blaze/elm/compiler/clinical_values_test.clj
```diff
@@ -136,7 +137,65 @@
...
+           [{:name "sys-def-120434"
+             :id "system-120411"
+             :version "version-120408"}
+            {:name "sys-def-115853"
+             :id "system-115911"
+             :version "version-115909"}]}}}]
+    (given (c/compile context #elm/concept [[["sys-def-120434" "code-115927"] ["sys-def-115853" "code-115928"]]])
+           type := Concept
+           [:codes 0 type] := Code
+           [:codes 0 :system] := "system-120411"
```
alexanderkiel, line 194: Also here.

modules/cql/test/blaze/elm/literal.clj
```diff
@@ -83,6 +83,17 @@
   {:type "CodeRef" :name name})
 
 
+;; 3.6. Concept
+(defn concept [[codes display]]
+  (cond->
+    {:type "Concept"
+     :codes (reduce #(conj %1 {:type "Code" :system
```
alexanderkiel, line 90: Can also be a `map`.

modules/cql/test/blaze/elm/literal.clj
```diff
@@ -83,6 +83,17 @@
...
+                               {:type "CodeSystemRef", :name (nth %2 0)}
```
alexanderkiel, line 91: Please use destructuring instead of `nth`.

modules/fhir-structure/resources/data_readers.clj
```diff
@@ -15,6 +15,7 @@
  fhir/dateTime blaze.fhir.spec.type/dateTime
  fhir/time blaze.fhir.spec.type/time
  fhir/code blaze.fhir.spec.type/code
+ fhir/concept blaze.fhir.spec.type/concept
```
alexanderkiel, line 18: This should be not here.

modules/cql/src/blaze/elm/concept.clj
```diff
@@ -0,0 +1,16 @@
+(ns blaze.elm.concept
```
alexanderkiel, line 1: There should be a `concept-spec` namespace under `test` similar to the `code-spec` namespace. This `concept-spec` namespace should also be required in the `clinical-values-test`.

alexanderkiel (changes requested):
Thanks for this good first part of the concept implementation. I see that the operators like equal will follow in a later PR. We should also add some integration tests under the `operation-measure-evaluate-measure` module in a later PR.

modules/cql/test/blaze/elm/concept_spec.clj
```diff
@@ -0,0 +1,15 @@
+(ns blaze.elm.concept-spec
+    (:require
```
alexanderkiel, line 2: Intendation is wrong

modules/cql/test/blaze/elm/concept_spec.clj
```diff
@@ -0,0 +1,15 @@
...
+    (:import
+      [blaze.elm.concept Concept]))
+
+
+(defn concept? [x]
+  (instance? Concept x))
+
+
+(s/fdef concept/to-concept
+  :args (s/coll-of (s/cat :system string? :version (s/nilable string?) :code string?))
```
alexanderkiel, line 14: The :args spec almost always [starts with `s/cat`](https://clojure.org/guides/spec#_specing_functions) so it will be `(s/cat :codes ...)`. For the "codes" spec, I would take `(s/coll-of code-spec/code?)` in order to test for a compiled code. For that to work, you need to require `[blaze.elm.code-spec :as code-spec]`.

modules/cql/test/blaze/elm/literal.clj
```diff
@@ -83,6 +83,18 @@
   {:type "CodeRef" :name name})
 
 
+;; 3.6. Concept
+(defn concept [[codes display]]
+  (cond->
+    {:type "Concept"
+     :codes (mapv (fn [x] (let [[sys code] x]
```
alexanderkiel, line 90: You can do the destructuring directly inside `fn`.

alexanderkiel (changes requested):

alexanderkiel (approved):

PR #844, opened 2022-11-27
Remove Conditional Patch

alexanderkiel:
It's an R5 property.


alexanderkiel:
This problem was introduced in #842 which was not released at this time. So it's no bug.

PR #860, opened 2022-12-08
Easy build without Gitlab CI/CD

lablans:
This adds a second Dockerfile to perform a full multi-stage build (Clojure, then Docker) without using Github CI/CD.


alexanderkiel:
I documented the local build process in favour of the redundant Dockerfile in #878.

PR #886, opened 2023-01-24
Add GitHub release when tag gets pushed

EmteZogaf:
The release step checks if the tag name is a valid semantic version number. It also adds the uberjar file as asset to the release.


alexanderkiel:
Did it in #891 without the version number checking.

PR #924, opened 2023-02-25
Implement Operation $graphql

alexanderkiel:
Currently only very view fields of Patient and Observation are supported.


hannesUlrich:
Hey @alexanderkiel 
I was looking into the $graphql and noticed that it is only partially implemented. Is the expansion of the schema in your roadmap? :)

alexanderkiel:
Hi @hannesUlrich,

we currently don't use GraphQL. So nothing substantial is planned here. But I can help you, if you like to expand the functionality.

PR #971, opened 2023-04-21
Temporarily Support Bundle Entries with Typeless Resources

alexanderkiel:
This should be reverted after it is clear why there are resources without a type.


alexanderkiel:
Was never really merged because the develop was force-pushed.

PR #989, opened 2023-05-08
Update Jackson to v2.15.0

alexanderkiel:
Had to set StreamReadConstraints.maxStringLen to 50 M in order to support at least our 8 MiB large binary resource test case. Added a download test in order to detect the config also for the CBOR parser used to read data from the resource store.


alexanderkiel:
This PR was actually merged. But GitHub doesn't notice it.

PR #1014, opened 2023-05-31
Use Jemalloc for RocksDB

alexanderkiel:
Using jemalloc solves memory fragmentation. I observed this problem while using block cache sizes above 1 GiB were RocksDB needed roughly 3 - 4 times the memory specified for the block cache.

See also: 
* http://smalldatum.blogspot.com/2018/04/myrocks-malloc-and-fragmentation-strong.html
* https://blog.cloudflare.com/the-effect-of-switching-to-tcmalloc-on-rocksdb-memory-use/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM eclipse-temurin:17-jre
 
 RUN apt-get update && apt-get upgrade -y && \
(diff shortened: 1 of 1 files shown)


alexanderkiel:
Closes: #1030

PR #1022, opened 2023-06-07
Fix Weight Calculation in Page Store

alexanderkiel:
I did not take into account that the PersistentVector/EMPTY_NODE is shared between vectors. So the size is a lot smaller.


alexanderkiel:
Closes: #1029

PR #1023, opened 2023-06-08
Increase Local Page Store Size

alexanderkiel:
We have also included the metric "blaze_page_store_estimated_size" to enable monitoring of the size. Additionally, we have decided to avoid generating tokens for searches that yield zero matches because such searches are fast and can result in a large number of useless tokens.


alexanderkiel:
Closes #1029

Issue #1268, opened 2023-11-13
Allow Handling of Large CQL Libraries


alexanderkiel:
Currently certain large CQL libraries can't be compiled because the translated ELM representation in JSON contains objects nested at a depth larger than 1000. There is a max nesting depth of 1000 in the `JsonFactory` used by the Jackson `ObjectMapper`. In this issue we like to increase that max nesting dept to 5000. If there are still reports of even larger queries not compilable, we would make the max depth a config option. But for now we don't like to expose an additional config option.

alexanderkiel:
We will not merge this. With the default JSON parser, we are able to handle up to 500 criteria in one named expression. We changed the CQL generation for the FDPG to output smaller named expressions. The ultimate goal should be to use value sets anyway. So there is no need to expend the number of possible criteria.

PR #1270, opened 2023-11-13
Allow Handling of Large CQL Libraries

alexanderkiel:
Currently certain large CQL libraries can't be compiled because the translated ELM representation in JSON contains objects nested at a depth larger than 1000. There is a max nesting depth of 1000 in the JsonFactory used by the Jackson ObjectMapper. In this issue we like to increase that max nesting dept to 5000. If there are still reports of even larger queries not compilable, we would make the max depth a config option. But for now we don't like to expose an additional config option.

Closes: #1268

### DIFF
--- a/modules/cql/src/blaze/cql_translator.clj
+++ b/modules/cql/src/blaze/cql_translator.clj
@@ -4,6 +4,7 @@
     [blaze.elm.spec]
     [jsonista.core :as j])
   (:import
+    [com.fasterxml.jackson.core JsonFactory StreamReadConstraints]
     [org.cqframework.cql.cql2elm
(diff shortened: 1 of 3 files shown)


alexanderkiel:
We will not merge this. With the default JSON parser, we are able to handle up to 500 criteria in one named expression. We changed the CQL generation for the FDPG to output smaller named expressions. The ultimate goal should be to use value sets anyway. So there is no need to expend the number of possible criteria.

alexanderkiel:
We will not merge this. With the default JSON parser, we are able to handle up to 500 criteria in one named expression. We changed the CQL generation for the FDPG to output smaller named expressions. The ultimate goal should be to use value sets anyway. So there is no need to expend the number of possible criteria.

PR #1336, opened 2023-12-09
Update Jetty to v12

alexanderkiel:
Jetty v12 requires JDK 17. This commit will remove the support for JDK 11.


alexanderkiel:
We are back to the standard Ring Jetty adapter in #2139.

PR #1449, opened 2024-01-28
Revert to Buffer using Code for resource-handle

alexanderkiel:
Because RocksDB v8.10.0 now allows to use heap byte buffers, we go back to the code that uses buffers for all iterator operations.


alexanderkiel:
Already done in #1450

PR #1621, opened 2024-04-15
Fix formats in manifest

### LINKED ISSUE #1619: Server metadata contains format "application/xml+json"
In Blaze version 0.25.0 the metadata contains

```
{
    ...
    "format": [
        "application/fhir+json"
        "application/xml+json"
    ],
    ...
}
```

Shouldn't the latter format be `application/fhir+xml` instead of `application/xml+json` which makes no sense?

pcvolkmer:
This will remove format "application/xml+json" and replaces it with "application/fhir+xml".

Also see http://hl7.org/fhir/R4/http.html#parameters for FHIR formats

closes #1619


pcvolkmer:
E-Tag header changed for manifest due to content change. Tests should  expect changed E-Tags.

pcvolkmer:
Frontend E2E tests failing due to playwright test timeouts. Add config option to set non default timeout (see: https://playwright.dev/docs/test-timeouts) might solve this issue as it occurs in other action runs as well (see: https://github.com/samply/blaze/actions/runs/8692051264/job/23836131479?pr=1618#step:22:13)

PR #1689, opened 2024-05-01
Make global repl explicit + docs improvements

allentiak:


DEVELOPMENT.md
```diff
@@ -49,11 +56,7 @@ In addition to the REPL development inside a single module, it's possible to run
 
 ## Style Guide
 
-The Clojure code in this project follows the [Clojure Style Guide][2].
-
-### Exceptions:
-
-* we use two empty lines between top-level forms
+The Clojure code in this project mainly follows the [Clojure Style Guide][2], except that we use 2 (two) empty lines between top-level forms.
```
alexanderkiel, line 59: We no longer use two empty lines between functions. In fact we use `cljfmt` now. So it would be good to mention `cljfmt` here.

alexanderkiel (changes requested):
Thanks for your contribution. Can you please rebase on develop and change the merge target to develop also?

DEVELOPMENT.md
allentiak: What about this?

"The Clojure code in this project mainly follows the [Clojure Style Guide][2], enforced by `cljfmt`. For more details, please check `cljfmt.edn`."

alexanderkiel: Perfect

allentiak:
It seems there is a token for running all the codecov tests that is missing...

I think there is an automated PR that fixes that...
https://github.com/samply/blaze/pull/1687

PR #1958, opened 2024-08-01
chore: add support for nREPL-based Clojure editors (buggy)

allentiak:
The -e parameter does not run

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ deps-list:
 	clojure -X:deps list
 
 repl:
-	clj -M:test -e "(require 'blaze.dev) (in-ns 'blaze.dev)" --repl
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -49,7 +49,7 @@ deps-list:
 	clojure -X:deps list
 
 repl:
-	clj -M:test -e "(require 'blaze.dev) (in-ns 'blaze.dev)" --repl
+	clj -M:test:repl/interactive -e "(require 'blaze.dev) (in-ns 'blaze.dev)"
```
allentiak, line 52: The original intention of the `-e` parameter was to automate the usage of the custom `blaze.dev` namespace. However, it does not run now... (I tried both with and without the `--repl` at the end).
Another option would be to switch to the standard `user` namespace, instead of the custom `blaze.dev` namespace.

allentiak:
Superseded by #2035

PR #2035, opened 2024-09-10
Add nREPL support

allentiak:
For now, big documentation changes will be part of a separate PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,8 +48,8 @@ deps-tree:
 deps-list:
 	clojure -X:deps list
 
-repl:
-	clj -M:test -e "(require 'blaze.dev) (in-ns 'blaze.dev)" --repl
+emacs-repl: prep
+	clj -M:test:emacs-repl
 
 cloc-prod-root:
 	cloc src
@@ -63,4 +63,4 @@ cloc-test: $(MODULES) cloc-test-root
 
 .PHONY: $(MODULES) lint-root lint prep test-root test test-coverage clean-root \
 	clean build-frontend build-ingress uberjar build-all outdated deps-tree \
-	deps-list cloc-prod cloc-test
+	deps-list emacs-repl cloc-prod cloc-test
--- a/deps.edn
+++ b/deps.edn
@@ -62,6 +62,30 @@
     {:git/tag "v0.10.5" :git/sha "2a21b7a"}}
    :ns-default build}
 
+  ;; For both toying at the terminal and/or use via any nREPL-aware IDE
+  ;; Automatically loads Emacs' middleware.
+  :emacs-repl
+  {:extra-deps
+   {nrepl/nrepl
+    {:mvn/version "1.3.0"}
+
+    ;; Emacs middleware below this line
+
+    cider/cider-nrepl
+    {:mvn/version "0.49.3"}
+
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -48,8 +48,8 @@ deps-tree:
 deps-list:
 	clojure -X:deps list
 
-repl:
-	clj -M:test -e "(require 'blaze.dev) (in-ns 'blaze.dev)" --repl
+repl: prep
```
alexanderkiel, line 51: the target should be also called `emacs-repl`

alexanderkiel: and please add the target to `.PHONY`

deps.edn
```diff
@@ -62,6 +62,19 @@
     {:git/tag "v0.10.5" :git/sha "2a21b7a"}}
    :ns-default build}
 
+  ;; For both toying at the terminal and/or use via any nREPL-aware IDE
+  ;; Automatically loads Emacs' middleware.
+  :emacs-repl
+  {:extra-deps {nrepl/nrepl                   {:mvn/version "1.3.0"}
```
alexanderkiel, line 68: you forgot to add line breaks after `:extra-deps` and each lib symbol.

deps.edn
```diff
@@ -62,6 +62,19 @@
...
+                ;; Emacs middleware
+                cider/cider-nrepl             {:mvn/version "0.49.3"}
+                refactor-nrepl/refactor-nrepl {:mvn/version "3.10.0"}
+                com.billpiel/sayid            {:mvn/version "0.1.0"}}
+   :main-opts  ["--eval" "(require 'blaze.dev) (in-ns 'blaze.dev)"
```
alexanderkiel, line 73: also add a line break after `:main-opts`

dev/blaze/dev.clj
```diff
@@ -34,11 +34,13 @@
 ;; Init Development
 (comment
   (init)
+  system
```
alexanderkiel, line 37: do you really need the system here? I mean it's available above.

alexanderkiel (changes requested):

dev/blaze/dev.clj
allentiak: You mean below?
That's why there are two comment blocks.

In this way, one can call the system var to verify the system is actually running in both cases without having to switch comment blocks.

alexanderkiel: Yes on line 22. I don't need to eval the system on regular basis. Why do you need it to tell that the system is running? If I doesn't;t run, there will be an exception during startup. Otherwise it runs.

deps.edn
```diff
@@ -62,6 +62,26 @@
     {:git/tag "v0.10.5" :git/sha "2a21b7a"}}
    :ns-default build}
 
+  ;; For both toying at the terminal and/or use via any nREPL-aware IDE
+  ;; Automatically loads Emacs' middleware.
+  :emacs-repl
+  {:extra-deps
+   {nrepl/nrepl
+    {:mvn/version "1.3.0"}
```
alexanderkiel, line 70: please add a newline after each version block.

alexanderkiel (changes requested):

PR #2043, opened 2024-09-12
Add `query-params` Validation

allentiak:


modules/operation-measure-evaluate-measure/Makefile
```diff
@@ -13,6 +13,12 @@ test: prep
 test-coverage: prep
 	clojure -M:test:coverage
 
+test-watch: prep
+	clojure -X:test:kaocha-watch --profile :watch
+
+repl: prep
```
alexanderkiel, line 19: should be called emacs-repl

modules/operation-measure-evaluate-measure/deps.edn
```diff
@@ -34,13 +34,33 @@
    {blaze/db-stub
     {:local/root "../db-stub"}}}
 
+  ;; For both toying at the terminal and editor support (via nREPL)
+  ;; Supports Emacs' refactor-nrepl and sayid
+  :repl
```
alexanderkiel, line 39: should be called emacs-repl

modules/operation-measure-evaluate-measure/deps.edn
```diff
@@ -34,13 +34,33 @@
...
+  {:extra-deps {nrepl/nrepl                   {:mvn/version "1.3.0"}
```
alexanderkiel, line 40: please use the indentation we discussed in the global deps.edn

modules/operation-measure-evaluate-measure/deps.edn
```diff
@@ -34,13 +34,33 @@
...
    {lambdaisland/kaocha
     {:mvn/version "1.91.1392"}}
 
    :main-opts ["-m" "kaocha.runner"]}
 
+  :kaocha-watch
+  {:extra-deps
+   {lambdaisland/kaocha
+    {:mvn/version "1.91.1392"}}
+   :exec-fn   kaocha.runner/exec-fn
```
alexanderkiel, line 59: I don't know about this new kind of execution config. But if that is now preferred over `:main-opts` we may use it everywhere?

modules/operation-measure-evaluate-measure/tests.edn
```diff
@@ -1,11 +1,13 @@
 #kaocha/v1
  #merge
   [{}
-   #profile {:ci {:reporter kaocha.report/documentation
-                  :color? false}
+   #profile {:ci       {:reporter kaocha.report/documentation
```
alexanderkiel, line 4: please don't align map values

alexanderkiel (changes requested):

modules/operation-measure-evaluate-measure/deps.edn
allentiak: I certainly agree in your consistency argument. That being said, I am not sure whether there is a preferred way at all or not - much less which one that would be.

So, I will leave this one open for now so I can research it further later on.

allentiak: For now, I will leave this change outside of this PR.
I will review this later.

allentiak: See https://github.com/samply/blaze/issues/2105

alexanderkiel (approved):

PR #2217, opened 2024-11-25
Fix Response Status Code for Failing XML Emits

allentiak:
In case that the generation of XML for responses failed, an OperationOutcome was already produced - but the status code was kept at 200. This was incorrect: in such a case that the server is unable to generate the XML output, the status code should be changed to 500.

### DIFF
--- a/modules/rest-util/src/blaze/middleware/fhir/output.clj
+++ b/modules/rest-util/src/blaze/middleware/fhir/output.clj
@@ -38,23 +38,31 @@
       (xml/emit (fhir-spec/unform-xml body) writer))
     (.toByteArray out)))
 
-(defn- generate-xml* [body]
-  (try
-    (generate-xml** body)
-    (catch Throwable e
-      (generate-xml** (handler-util/operation-outcome (ba/anomaly e))))))
+(defn- generate-xml-error [e]
+  (-> e
+      ba/anomaly
+      handler-util/operation-outcome
+      generate-xml**))
 
-(defn- generate-xml [body]
+(defn- generate-xml* [response]
+  (try
+    (update response :body generate-xml**)
+    (catch Throwable e
+      (assoc response
+             :body (generate-xml-error e)
+             :status 500))))
+
+(defn- generate-xml [response]
   (log/trace "generate XML")
   (with-open [_ (prom/timer generate-duration-seconds "xml")]
-    (generate-xml* body)))
+    (generate-xml* response)))
 
 (defn- encode-response-json [{:keys [body] :as response} content-type]
   (cond-> response body (-> (update :body generate-json)
                             (ring/content-type content-type))))
 
 (defn- encode-response-xml [{:keys [body] :as response} content-type]
-  (cond-> response body (-> (update :body generate-xml)
+  (cond-> response body (-> generate-xml
                             (ring/content-type content-type))))
 
 (defn- format-key [format]
--- a/modules/rest-util/test/blaze/middleware/fhir/output_test.clj
+++ b/modules/rest-util/test/blaze/middleware/fhir/output_test.clj
@@ -138,7 +138,9 @@
 
   (testing "failing XML emit"
     (given (call (special-resource-handler {:fhir/type :fhir/Patient :id "0" :gender #fhir/code"foo\u001Ebar"}) {:headers {"accept" "application/fhir+xml"}})
+      :status := 500
       [:headers "Content-Type"] := "application/fhir+xml;charset=utf-8"
+      [:body parse-xml :fhir/type] := :fhir/OperationOutcome
       [:body parse-xml :issue 0 :diagnostics] := "Invalid white space character (0x1e) in text to output (in xml 1.1, could output as a character entity)")))
 
 (deftest not-acceptable-test


modules/rest-util/src/blaze/middleware/fhir/output.clj
```diff
@@ -32,29 +32,37 @@
   (with-open [_ (prom/timer generate-duration-seconds "json")]
     (fhir-spec/unform-json body)))
 
-(defn- generate-xml** [body]
+(defn- xml-byte-array [body]
```
alexanderkiel, line 35: `xml-byte-array` is, in many aspects, not a good name. First it doesn't convert some XML into a byte-array, second I don't thing the aspect that the return type is a byte-array has such an importance here, third, there is already an established name for generating an XML response body.

Also the convention to postfix function names with stars that need to be broken up in vertical terms, is quite idiomatic in Clojure. All the variants of `generate-xml` actually generate XML but the first one wraps metric collection and the second one wraps exception handling.

modules/rest-util/src/blaze/middleware/fhir/output.clj
```diff
@@ -32,29 +32,37 @@
...
   (let [out (ByteArrayOutputStream.)]
     (with-open [writer (io/writer out)]
       (xml/emit (fhir-spec/unform-xml body) writer))
     (.toByteArray out)))
 
-(defn- generate-xml* [body]
+(defn- e->xml-byte-array [e]
```
alexanderkiel, line 41: if the body if this function doesn't fit into the line it is called, than please name if `generate-xml-error`

modules/rest-util/test/blaze/middleware/fhir/output_test.clj
```diff
@@ -136,9 +136,11 @@
       [:headers "Content-Type"] := nil
       :body := nil))
 
-  (testing "failing XML emit"
+  (testing "invalid XML"
```
alexanderkiel, line 139: "failing XML emit" was better, because there is no invalid XML, there is a Patient in the internal FHIR representation that can't be emitted as XML, but it's no XML in the first place. so the function from internal FHIR representation to XML is only partial.

alexanderkiel (changes requested):
Your solution is functionally perfect. It's only the naming I have to criticise.

modules/rest-util/test/blaze/middleware/fhir/output_test.clj
allentiak: I see.
Then, what about "failed conversion (XML)", or similar?

modules/rest-util/src/blaze/middleware/fhir/output.clj
allentiak: I see your point. I've seen the star convention you mention in other portions of the code, it makes sense.

I only tried to make the name different in this case (my choice probably being not the best guess, as you explained above) because, within the whole `generate-xml[*,**,***]` function chain, this one was the only one that still used the body, and not the whole response. Having two functions whose name only differs by one star with two different interfaces seems error-prone, specially for someone who, in the not-so-near future comes and revisits that piece of code.

Or am I overthinking this?

modules/rest-util/src/blaze/middleware/fhir/output.clj
allentiak: But it's not generating any error. This function is simply a pipeline that *converts* an exception into an xml-byte-array (or however we end up calling the other function's result).

allentiak: I can certainly rename it to `e->whatever`, but it's a pipeline function. That's why I used [the `->` convention](https://guide.clojure.style/#naming-conversion-functions).
Or would you suggest doing this differently?

modules/rest-util/src/blaze/middleware/fhir/output.clj
alexanderkiel: I see your point with the different interface of the functions. Weighting all options, I think, we should keep the original name.

modules/rest-util/test/blaze/middleware/fhir/output_test.clj
alexanderkiel: Can you please keep the doc. It was already good.

modules/rest-util/src/blaze/middleware/fhir/output.clj
alexanderkiel: The -> convention is about a conversion function, not about a pipeline function. Here you can say that an error is converted into a byte array containing the error in XML format. But I like `generate-xml` better than just saying that we convert some data type into another data type. Generating is like the inverse of parsing. 

And please don't try to change everything, if the task is just to add a 500 status code.

alexanderkiel (changes requested):
If you please could change the commit message into something like this:

>Fix Response Status Code for Failing XML Emits
>
>In case the generation of XML for responses failed, an OperationOutcome was
>already produced, but the status code was kept at 200. However in such a
>case that the server was unable to generate the XML output, the status
>code has to be 500.

I would be perfect.

alexanderkiel (approved):

Issue #2126, opened 2024-10-18
Implement Binary Upload


allentiak:
Specification:

- When binary data is written to the server ([create](https://hl7.org/fhir/R4/http.html#create)/[update](https://hl7.org/fhir/R4/http.html#update) - POST or PUT), the data is accepted as is and treated as the content of a Binary, including when the content type is "application/fhir+xml" or "application/fhir+json", except for the special case where the content is actually a Binary resource.

Scaffolding:

* Check the `blaze.middleware.fhir.resource` middleware. There, we also need a `wrap-binary-resource` function that has to allow non-fhir content-types and will create a Binary resource on the fly using the body as `:data` (base64 encoded, of course) and adding the `content-type` as `:contentType`.

* At the end, both moddleware functions, the already existing `wrap-resource` and the new `wrap-binary-resource` have to emit a FHIR resource to the next middleware/handler.

* We will also need integration tests for this feature. In order to implement them, we can use the `openssl` command like in `.github/scripts/generate-large-binary-resource.sh` to generate such a large 8 MiB random content for the integration test. We should be able to upload and download that content and it should be equal afterwards.

Part of #2009.

PR #2255, opened 2024-12-13
Improve naming (preparation for #2126)

allentiak:


.github/scripts/read-binary-content-via-json-invalid.sh
```diff
@@ -0,0 +1,90 @@
...
+echo "  - via JSON, with ID: $ID_VIA_JSON"
+
+
+# Retrieve the Binary resource, and Base64 encode it so it can be safely handled by Bash (JSON)
+BASE64_ENCODED_BINARY_RESOURCE_VIA_JSON=$(curl -s -H 'Accept: application/pdf' "$BASE/Binary/$ID_VIA_JSON" | base64 | tr -d '\n')
+
+
+echo "Binary data retrieved. Verifying content... (JSON version)"
+
+if [ "$DATA" = "$BASE64_ENCODED_BINARY_RESOURCE_VIA_JSON" ]; then
```
alexanderkiel, line 85: But if this is true, no error was generated.

At the end, after thinking about it, it should not be possible to upload faulty data. So it will be not easy to test the error generation. We had to manipulate the data inside the database. So I would say, we should skip this test entirely.

alexanderkiel (changes requested):

modules/rest-util/test/blaze/middleware/fhir/output_test.clj
```diff
@@ -214,12 +214,12 @@
           [:headers "Content-Type"] := nil
           :body := nil)))
 
-    (testing "failing binary emit (with invalid data)"
-      (given (call (binary-resource-handler-200 {:content-type "application/pdf" :data "MTANjECg=="}) {:headers {"accept" "text/plain"}})
+    (testing "failing binary emit (with invalid data, using the JSON default)"
+      (given (call (binary-resource-handler-200 {:content-type "application/fhir+json" :data "MTANjECg=="}) {:headers {"accept" "text/plain"}})
```
alexanderkiel, line 218: The content-type within the Binary resource is still `application/pdf`.

modules/rest-util/src/blaze/middleware/fhir/output.clj
```diff
@@ -107,15 +107,15 @@
     "xml" :fhir+xml
     nil))
 
-(defn- request-format
+(defn- accept-header-format
```
alexanderkiel, line 110: This is not only about the `Accept` header. Also the `_format` query param is considered. That's why I called the function `request-format`.

alexanderkiel (changes requested):

alexanderkiel (approved):

PR #2256, opened 2024-12-13
Clearly separate FHIR and non-FHIR middleware (preparation for #2126)

allentiak:


alexanderkiel:
We have agreed to not use the `non-fhir` namespaces.

allentiak:
Thanks for closing this.

PR #2306, opened 2025-01-09
Reorganize and Improve tests

allentiak:


alexanderkiel:
I see no value in adding `application/pdf` and another obscure media type to the tests.

PR #2307, opened 2025-01-09
Fix Binary Resource Generation Logic

allentiak:


alexanderkiel:
I close this in favour of #2308 because it dosen't solve the issue.

PR #2321, opened 2025-01-11
Update vite to v6

alexanderkiel:


alexanderkiel:
Will not work until vitest v3 is released.

Issue #2292, opened 2025-01-04
Pass Validation Simple Terminology Tests


alexanderkiel:
Pass all [test cases](https://build.fhir.org/ig/HL7/fhir-tx-ecosystem-ig/testcases.html) starting with `validation-simple`.

PR #2326, opened 2025-01-14
Pass Validation Simple Terminology Tests

alexanderkiel:
Closes: #2292


alexanderkiel:
Passing the simple validation tests, is not so straight forward as I thought. So I will close that PR and concentrate more on features we actually need for validation.

PR #2331, opened 2025-01-21
Implement Binary Upload

### LINKED ISSUE #2126: Implement Binary Upload

allentiak:


modules/rest-api/src/blaze/rest_api/routes.clj
```diff
@@ -98,134 +98,134 @@
   (when-let
    [{:blaze.rest-api.resource-pattern/keys [interactions]}
     (u/resolve-pattern resource-patterns structure-definition)]
-    (cond->
```
allentiak, line 101: According to `parinfer`, the `(cond->` should be aligned with the `[` from the `when-let`.

alexanderkiel (changes requested):

.github/scripts/write-binary-content-via-raw-data.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# 8 MiB of random data, base64 encoded (8388608 bytes), kept in memory
+DATA="$(openssl rand -base64 8388608 | tr -d '\n')"
+# Also keep a raw binary version in memory for direct upload
+RAW_DATA="$(echo "$DATA" | base64 -d)"
+
+echo "Testing direct binary upload and download..."
+
+# Create Binary resource via direct binary upload
+ID_VIA_BINARY=$(curl -s -H 'Content-Type: application/octet-stream' --data-binary <(echo -n "$RAW_DATA") "$BASE/Binary" | jq -r '.id')
```
alexanderkiel, line 16: is the echo piping necessary? a normal string will not work with 8 MiB?

alexanderkiel, line 16: Just `ID` will do it in this script were you use only binary upload

.github/scripts/write-binary-content-via-raw-data.sh
```diff
@@ -0,0 +1,43 @@
...
+
+if [ "$DATA" = "$BASE64_ENCODED_BINARY_RESOURCE_VIA_DIRECT" ]; then
+    echo "✅ Direct Binary: Successfully verified 8 MiB binary content integrity"
+else
+    echo "🆘 Direct Binary: Content verification failed"
+    exit 1
+fi
+
+# Verify sizes (all in memory)
+ORIGINAL_SIZE=$(echo -n "$RAW_DATA" | wc -c)
```
alexanderkiel, line 31: I mean you already have compared the data directly, why do you like to compare the length afterwards?

modules/rest-api/src/blaze/rest_api/routes.clj
```diff
@@ -103,6 +107,8 @@
       {:fhir.resource/type name}
       [""
        (cond-> {:name (keyword name "type")}
+         (= name "Binary")
```
alexanderkiel, line 110: not needed at type level

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -48,10 +49,10 @@
      :fhir/issue "structure"
      :fhir/operation-outcome "MSG_JSON_OBJECT")))
 
-(defn- resource-request-json [{:keys [body] :as request}]
+(defn- resource-as-json [{:keys [body] :as request}]
```
alexanderkiel, line 52: is there a reason to change that name?

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -48,10 +49,10 @@
...
   (if body
-    (when-ok [x (parse-json body)
-              resource (conform-json x)]
+    (when-ok [b (parse-json body)
```
alexanderkiel, line 54: also here, why do you change that name?

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -89,39 +90,78 @@
...
+(defn- resource-as-xml [{:keys [body] :as request}]
   (if body
     (when-ok [resource (parse-and-conform-xml body)]
       (assoc request :body resource))
     (ba/incorrect "Missing HTTP body.")))
 
+(defn- get-binary-data [body]
+  (with-open [_ (prom/timer parse-duration-seconds "binary")
+              ^InputStream input body]
+    (slurp input)))
```
alexanderkiel, line 102: base64 encoding is missing

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -89,39 +90,78 @@
...
+    (cond
+      (json-request? content-type) (resource-as-json request)
+      (xml-request? content-type) (resource-as-xml request)
+      :else
+      (resource-as-binary-data request))
+    (ba/incorrect "Expected Content-Type header for binary resources.")))
 
 (defn wrap-resource
-  "Middleware to parse a resource from the body according the content-type
+  "Middleware to parse a FHIR resource from the body according the content-type
```
alexanderkiel, line 140: we should discuss such doc changes before

alexanderkiel (changes requested):
Please also look if we still need the `big-binary-test` in build.yml.

.github/scripts/write-binary-content-via-raw-data.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash -e
+
+# This script creates a large binary resource (8 MiB) and verifies that its binary content
+# can be read correctly via both JSON/XML wrappers and direct binary upload.
+
+BASE="http://localhost:8080/fhir"
+
+# 8 MiB of random data, base64 encoded (8388608 bytes), kept in memory
+DATA="$(openssl rand -base64 8388608 | tr -d '\n')"
```
alexanderkiel, line 9: you don't upload base64 encoded data, it has to work with arbitrary binary data

alexanderkiel (changes requested):

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -95,6 +99,25 @@
       (assoc request :body resource))
     (ba/incorrect "Missing HTTP body.")))
 
+(def ^:private ^Base64$Encoder b64-encoder
+  (.withoutPadding (Base64/getUrlEncoder)))
```
alexanderkiel, line 103: why URL encoder and why without padding? The encoder has to fit the `(.decode (Base64/getDecoder) ...)` from the output. Does FHIR say something about Url encoding and padding? Are we wrong with the output?

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -95,6 +99,25 @@
...
+
+(defn- get-binary-data [body]
+  (with-open [_ (prom/timer parse-duration-seconds "binary")
+              ^InputStream input body]
+    (.encodeToString b64-encoder (.readAllBytes input))))
```
alexanderkiel, line 108: I would put this including the encoder in a function with all the type hints. especially the binding of body to input in with-open is not necessary. it would be possible to just tag body with InputStream. but I usually try to encapsulate all type hints in a simple function

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -95,6 +99,25 @@
...
+
+(defn- resource-request-binary-data [{:keys [body headers] :as request}]
+  (if body
+    (when-ok [b64-encoded-data (get-binary-data body)]
```
alexanderkiel, line 112: I don't see that an anomaly can be created here

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -95,6 +99,25 @@
...
+      (let [content-type (get headers "content-type")]
+        (assoc request :body
+               {:fhir/type :fhir/Binary
+                :resourceType "Binary"
```
alexanderkiel, line 116: not necessary, the internal format uses only :fhir/type

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -109,19 +132,43 @@
                     :http/status 415))
     (if (str/blank? (slurp (:body request)))
       (assoc request :body nil)
-      (ba/incorrect "Content-Type header expected, but is missing."))))
+      (ba/incorrect "Expected Content-Type header for FHIR resources."))))
+
+(defn- binary-resource-request [request]
+  (if-let [content-type (request/content-type request)]
+    (cond
+      (json-request? content-type) (resource-request-json request)
```
alexanderkiel, line 140: you can't use that json-request? function, because requesting application/json should return the binary content only _format and application/fhir+json should return the resource

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -109,19 +132,43 @@
...
+      (xml-request? content-type) (resource-request-xml request)
+      :else
+      (resource-request-binary-data request))
+    (ba/incorrect "Expected Content-Type header for binary resources.")))
```
alexanderkiel, line 144: my text from the resource-request function is: Content-Type header expected, but is missing.

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -109,19 +132,43 @@
...
 
 (defn wrap-resource
   "Middleware to parse a resource from the body according the content-type
   header.
 
-  Updates the :body key in the request map on successful parsing and conforming
-  the resource to the internal format.
+  If the resource is successfully parsed and conformed to the internal format,
```
alexanderkiel, line 150: no change please

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -109,19 +132,43 @@
...
   (fn [request]
     (if-ok [request (resource-request request)]
       (handler request)
       ac/completed-future)))
+
+(defn wrap-binary-data
+  "Middleware to parse binary data from the body according the content-type
+  header.
+
+  If the resource is successfully parsed and conformed to the internal format,
```
alexanderkiel, line 165: please use my wording, even if it's bad English, we can improve that later

alexanderkiel (changes requested):
I did not look at the unit tests, will do that after you worked on the current issues

modules/rest-util/test/blaze/middleware/fhir/resource_test.clj
```diff
@@ -15,33 +16,44 @@
...
-(defn wrap-error [handler]
+(defn- wrap-error [handler]
   (fn [request]
     (-> (handler request)
         (ac/exceptionally handler-util/error-response))))
 
-(def resource-handler
-  "A handler which just returns the :body from the request."
+(def ^:private resource-handler
+  "A handler which just returns the `:body` from a non-binary resource request."
```
alexanderkiel, line 34: I don't use backticks for keywords. I think, I saw this in Clojure core docs. Keywords already stand out in text.

modules/rest-util/test/blaze/middleware/fhir/resource_test.clj
```diff
@@ -193,6 +205,41 @@
               :body (input-stream (str "<Binary xmlns=\"http://hl7.org/fhir\"><data value=\"" (apply str (repeat (* 8 1024 1024) \a)) "\"/></Binary>"))})
       fhir-spec/fhir-type := :fhir/Binary)))
 
+(deftest binary-test
+  (testing "when sending FHIR-wrapped Binary data"
+    (testing "both handlers should return the same for both wrappers (JSON and XML)"
+      (let [b64-encoded-binary-data "MTA1NjE0Cg==" ;; raw data: 105614
+            binary-resource-data (type/base64Binary b64-encoded-binary-data)
+            binary-resource-content-type (type/code "text/plain")]
```
alexanderkiel, line 213: content type should be in string form here not FHIR code, because you use it in string concatenation, just wrap it into a FHIR code later in the given block

alexanderkiel (changes requested):

allentiak (commented):
Could you please review this again?
My only doubt is whether we should keep the debugging code in the integration test. I think it may be useful in the future...

modules/rest-util/test/blaze/middleware/fhir/resource_test.clj
```diff
@@ -190,21 +219,57 @@
...
+                [fhir-content-type resource-string-representation]
+                [["application/fhir+json;charset=utf-8" (resource-as-json b64-encoded-binary-data binary-resource-content-type)]
+                 ["application/fhir+xml;charset=utf-8" (resource-as-xml b64-encoded-binary-data binary-resource-content-type)]]]
+          (let [closed? (atom false)]
+            (given @(handler
+                     {:headers {"content-type" fhir-content-type}
+                      :body (string-input-stream resource-string-representation closed?)})
+              {:fhir/type :fhir/Binary
+               :contentType (type/code binary-resource-content-type)
+               :data binary-resource-data})
```
alexanderkiel, line 241: you can inline that binding, because it's only used once and should fit in one line

modules/rest-util/test/blaze/middleware/fhir/resource_test.clj
```diff
@@ -190,21 +219,57 @@
...
+            (is closed?))))))
+
+  (testing "when sending raw binary resource handling"
+    (let [small-data (bytes-of-random-binary-data 16)
+          large-data (bytes-of-random-binary-data (* 8 1024 1024))]
+      (doseq [[content-type raw-data]
+              [["text/plain" small-data]
+               ["application/octet-stream" large-data]]]
+        (let [closed? (atom false)
+              encoded-data (encode-binary-data raw-data)]
```
alexanderkiel, line 251: I think, this can also be inlined

modules/rest-util/test/blaze/middleware/fhir/resource_test.clj
```diff
@@ -15,47 +16,75 @@
...
+(defn- string-input-stream
   ([^String s]
    (ByteArrayInputStream. (.getBytes s StandardCharsets/UTF_8)))
   ([^String s closed?]
    (proxy [ByteArrayInputStream] [(.getBytes s StandardCharsets/UTF_8)]
      (close []
        (reset! closed? true)))))
 
+(defn- binary-input-stream
+  ([^bytes data]
```
alexanderkiel, line 55: as I see it, the form without closed? isn;t used - right?

modules/rest-util/test/blaze/middleware/fhir/resource_test.clj
```diff
@@ -190,21 +219,57 @@
...
+              [["text/plain" small-data]
+               ["application/octet-stream" large-data]]]
+        (let [closed? (atom false)
+              encoded-data (encode-binary-data raw-data)]
+          (given @(binary-resource-handler
+                   {:headers {"content-type" content-type}
+                    :body (binary-input-stream raw-data closed?)})
+            {:fhir/type :fhir/Binary
+             :resourceType "Binary"
+             :contentType content-type
```
alexanderkiel, line 257: please wrap that in type/code

modules/rest-util/test/blaze/middleware/fhir/resource_test.clj
```diff
@@ -190,21 +219,57 @@
...
+
+  (testing "when sending raw binary resource handling"
+    (let [small-data (bytes-of-random-binary-data 16)
+          large-data (bytes-of-random-binary-data (* 8 1024 1024))]
+      (doseq [[content-type raw-data]
+              [["text/plain" small-data]
+               ["application/octet-stream" large-data]]]
+        (let [closed? (atom false)
+              encoded-data (encode-binary-data raw-data)]
+          (given @(binary-resource-handler
```
alexanderkiel, line 252: given is not used right here. you are not asserting anything this way

modules/rest-util/test/blaze/middleware/fhir/resource_test.clj
```diff
@@ -190,21 +219,57 @@
...
+    (testing "both handlers should return the same for both wrappers (JSON and XML)"
+      (let [b64-encoded-binary-data "MTA1NjE0Cg=="
+            binary-resource-data (type/base64Binary b64-encoded-binary-data)
+            binary-resource-content-type "text/plain"]
+        (doseq [handler [resource-handler binary-resource-handler]
+                [fhir-content-type resource-string-representation]
+                [["application/fhir+json;charset=utf-8" (resource-as-json b64-encoded-binary-data binary-resource-content-type)]
+                 ["application/fhir+xml;charset=utf-8" (resource-as-xml b64-encoded-binary-data binary-resource-content-type)]]]
+          (let [closed? (atom false)]
+            (given @(handler
```
alexanderkiel, line 236: given is not used right here. you are not asserting anything this way

alexanderkiel (changes requested):

modules/rest-util/test/blaze/middleware/fhir/resource_test.clj
allentiak: Actually, it is. (In the JSON and the XML tests.)
Do you want me to change those tests to use the form with `closed?`?
I could do that in a new PR...

modules/rest-util/src/blaze/middleware/fhir/resource.clj
```diff
@@ -109,19 +130,43 @@
                     :http/status 415))
     (if (str/blank? (slurp (:body request)))
       (assoc request :body nil)
-      (ba/incorrect "Content-Type header expected, but is missing."))))
+      (ba/incorrect "Missing Content-Type header for FHIR resources."))))
+
+(defn- binary-resource-request [request]
+  (if-let [content-type (request/content-type request)]
+    (cond
+      (str/starts-with? content-type "application/fhir+json") (resource-request-json request)
```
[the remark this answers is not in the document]
alexanderkiel: https://hl7.org/FHIR/binary.html#rest

modules/rest-util/test/blaze/middleware/fhir/resource_test.clj
alexanderkiel: I don't wee the usage. Please point me to the line.

alexanderkiel (changes requested):

alexanderkiel (approved):

PR #2347, opened 2025-01-24
Improve Binary Download tests

allentiak:
It was not clear that, no matter what is the requested content type, as long as it is a non-FHIR one, the actual content type will be that of the resource.


alexanderkiel (changes requested):
Your discovery is right, but by using such media types we would look like idiots. It would be possible to use `text/whatever-you-want` or just keep the current tests and move one.

alexanderkiel (approved):

PR #2521, opened 2025-03-28
Move Conditional Delete Docs into Type Interaction

### LINKED ISSUE #2495: Move Conditional Delete Docs into Type Interaction
Conditional Delete is a type interaction not an instance interaction. See: https://www.hl7.org/fhir/http.html#cdelete

mmwaikar:
closes #2495

- create blaze-icon.svg from blaze-logo.svg
- create blaze-icon.ico from the blaze-icon.svg
- show this icon as favicon (in config.ts)
- create delete-type.md for type level conditional delete


docs/.vitepress/config.ts
```diff
@@ -3,6 +3,7 @@ import { defineConfig } from 'vitepress';
 export default defineConfig({
 	title: 'Blaze',
 	description: 'Blaze FHIR Server Documentation',
+	head: [['link', { rel: 'icon', href: '/blaze-icon.ico' }]],
```
alexanderkiel, line 6: Please take/copy the favicon from `modules/frontend/static/favicon.ico` and also name it so.

docs/.vitepress/config.ts
```diff
@@ -78,6 +79,7 @@ export default defineConfig({
 								collapsed: true,
 								items: [
 									{ text: 'Create', link: '/api/interaction/create' },
+									{ text: 'Delete', link: '/api/interaction/delete-type' },
```
alexanderkiel, line 82: Please also add a nav entry into `docs/api.md`.

alexanderkiel (changes requested):
Do we need the blaze-icon.svg?

docs/.vitepress/config.ts
mmwaikar: I tried all 3 files - `favicon.ico`, `favicon.svg` & `favicon-96x96.png` - the first one does not show up, the last 2 work, hence I am keeping only the `svg` file.

I've also deleted the `blaze-icon.svg` file which I had created - I kept it only because we could generate the `ico` file based on the `svg` file.

alexanderkiel (approved):

docs/.vitepress/config.ts
alexanderkiel: We might have to revisit that. But I'll leave it because the issue was not about the favicon at all.

Issue #2620, opened 2025-04-30
Add a Note that /etc/hosts Entry is Required


alexanderkiel:
In the [Full Standalone Deployment](https://samply.github.io/blaze/deployment/full-standalone.html) documentation, the hostnames `blaze.localhost` and `keycloak.localhost` are used in order to differentiate between Blaze and Keycloak. Browsers except Safari and `curl` automatically support subdomains of `localhost` but `blazectl` isn't. So an `/etc/hosts` entry is necessary.

PR #2622, opened 2025-04-30
Add a Note that /etc/hosts Entry is Required

alexanderkiel:
The note on Safari was removed, because with the `/etc/hosts` entry, Safari also works.

Closes: #2620

### DIFF
--- a/docs/deployment/full-standalone.md
+++ b/docs/deployment/full-standalone.md
@@ -30,6 +30,15 @@ The `docs/deployment/full-standalone` directory contains a Docker Compose file w
 > [!NOTE]
 > This example setup is for a developer's machine (on localhost) and things may be different on a real virtual machine.
 
+Create two entries in your `/etc/hosts` (`C:\Windows\System32\Drivers\Etc\hosts` on Windows):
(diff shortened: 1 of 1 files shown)


docs/deployment/full-standalone.md
```diff
@@ -30,6 +30,15 @@ The `docs/deployment/full-standalone` directory contains a Docker Compose file w
...
 > This example setup is for a developer's machine (on localhost) and things may be different on a real virtual machine.
 
+Create two entries in your `/etc/hosts` (`C:\Windows\System32\Drivers\Etc\hosts` on Windows):
+
+```
+127.0.0.1       blaze.localhost
+127.0.0.1       keycloak.localhost
+```
+
+That subdomains of `localhost` are used as virtual hosts of the nginx ingress container. They ensure that Keycloak and Blaze are separated properly. Other hostnames can be used, but have to be changed everywhere in this setup.
```
mmwaikar, line 40: These subdomains (instead of That)

mmwaikar (approved):
Approved.

PR #2681, opened 2025-05-21
Implement BCP 47 Code System

### LINKED ISSUE #2505: Implement BCP 47 Code System
The code system [urn:ietf:bcp:47](https://terminology.hl7.org/6.2.0/CodeSystem-v3-ietf3066.html) has no FHIR representation and so needs to be implemented directly.

The BCP-47 spec can be found [here](https://www.rfc-editor.org/info/bcp47). The implementation should be analogous to that of [BCP-13](https://github.com/samply/blaze/blob/main/modules/terminology-service/src/blaze/terminology_service/local/code_system/bcp_13.clj)

mmwaikar:
closes #2505

- show script and region with locale
- add missing tests for bcp-13

### DIFF
--- a/modules/terminology-service/test/blaze/terminology_service/local_test.clj
+++ b/modules/terminology-service/test/blaze/terminology_service/local_test.clj
@@ -111,6 +111,19 @@
    :blaze.test/fixed-rng-fn {}
    ::ts-local/graph-cache {}))
 
+(def bcp-47-config
+  (assoc
+   mem-node-config
+   ::ts/local
+   {:node (ig/ref :blaze.db/node)
+    :clock (ig/ref :blaze.test/fixed-clock)
+    :rng-fn (ig/ref :blaze.test/fixed-rng-fn)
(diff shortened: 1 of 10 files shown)


modules/terminology-service/test/blaze/terminology_service/local_test.clj
```diff
@@ -889,6 +926,64 @@
           [(parameter "system") 0 :value] := #fhir/uri"system-115910"
           [(parameter "version") 0 :value] := #fhir/string"version-124939")))))
 
+(deftest code-system-validate-code-bcp-47-test
```
alexanderkiel, line 929: please put this after the bcp-13 test

alexanderkiel (changes requested):

modules/terminology-service/test/blaze/terminology_service/local_test.clj
mmwaikar: Done.

alexanderkiel (approved):

Issue #2682, opened 2025-05-21
Support Summary Mode Search Type


alexanderkiel:
This time we can support summary for all resource types instead of only CodeSystem and ValueSet before in #2527.

PR #2685, opened 2025-05-22
Support Summary Mode Search Type

alexanderkiel:
Closes: #2682


modules/interaction/test/blaze/interaction/search_type_test.clj
knoppiks, line 1: Shouldn't we check for `_summary=false`, too?

knoppiks (changes requested):

modules/interaction/test/blaze/interaction/search_type_test.clj
alexanderkiel: done

knoppiks (approved):

PR #2762, opened 2025-06-24
Improve String Concatenation Performance

alexanderkiel:
The function clojure.core/str that is usually used for string concatenation has a performance problem which originates in missing arities. The str function has only the arities 0, 1, and variadic. That means that concatenating two strings will use the variadic arity which will put the second argument into a list before the variadic version is called. The implementation uses a StringBuilder that is initialized with the first string while all other strings are appended. Doing so will result in a StringBuilder eventually having to resize it's internal byte array. On the other hand, if the str function had multiple arities, individual known strings can be concatenated.


modules/util/test/blaze/util_test.clj
```diff
@@ -0,0 +1,54 @@
...
+
+(deftest strip-leading-slash-test
+  (satisfies-prop 10000
+    (prop/for-all [s gen/string]
+      (not (str/starts-with? (u/strip-leading-slashes s) "/")))))
+
+(deftest available-processors-test
+  (is (pos-int? (u/available-processors))))
+
+(deftest str-test
```
knoppiks, line 39: I expected to see the implementation compared to original `clojure.core/str`. Comparing against some constants might be okay - but using `clojure.core/str`as target would enable usage of property based testing.

knoppiks (changes requested):

modules/util/test/blaze/util_test.clj
alexanderkiel: Done

knoppiks (approved):

PR #2780, opened 2025-06-28
Exclude Artefact si.uom/si-quantity

alexanderkiel:
I have to exclude the artefact si.uom/si-quantity because it's classes are already contained in si.uom/si-units.

See: https://github.com/unitsofmeasurement/uom-systems/issues/206

### DIFF
--- a/modules/cql/deps.edn
+++ b/modules/cql/deps.edn
@@ -42,7 +42,8 @@
   {:mvn/version "2.2"}
 
   systems.uom/systems-ucum
-  {:mvn/version "2.2"}}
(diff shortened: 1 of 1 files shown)


keilw:
Thanks, looks like there is a problem with the module-info, because both modules USE it, but none of them declares or exposes it.
https://stackoverflow.com/questions/46537294/whats-the-difference-between-requires-and-requires-static-in-module-declaration explains, why this might happen. I changed the declaration of si.quantity to static now everywhere in uom-systems. That seems to work, but if it does not have any negative side-effect, I'll also include your PR.

PR #2792, opened 2025-07-02
Test Queries with Multiple Codes and Patients

alexanderkiel:
Before we had performance and functional tests for queries with a single code like a laboratory observation with a certain LOINC code over multiple patients.

This commit adds one functional test and a performance test for queries with multiple codes in addition to multiple patients.

For the functional download tests with query, I added an additional check of resource uniqueness.


.github/scripts/download-resources-query-sort.sh
```diff
@@ -30,7 +30,7 @@ else
   exit 1
 fi
 
-if [ "$(diff "$FILE_NAME_PREFIX-get.ndjson" "$FILE_NAME_PREFIX-post.ndjson")" = "" ]; then
+if [ -z "$(diff "$FILE_NAME_PREFIX-get.ndjson" "$FILE_NAME_PREFIX-post.ndjson")" ]; then
```
knoppiks: `diff` already returns a non-zero exit code when the files are not equal. Therefor we could further simplify this using
```suggestion
if diff "$FILE_NAME_PREFIX-get.ndjson" "$FILE_NAME_PREFIX-post.ndjson >/dev/null; then
```

.github/scripts/download-resources-query.sh
```diff
@@ -43,7 +52,7 @@ else
   exit 1
 fi
 
-if [ "$(diff "$FILE_NAME_PREFIX-get.ndjson" "$FILE_NAME_PREFIX-post.ndjson")" = "" ]; then
+if [ -z "$(diff "$FILE_NAME_PREFIX-get.ndjson" "$FILE_NAME_PREFIX-post.ndjson")" ]; then
```
knoppiks: Same as in `download-resources-query-sort.sh`

.github/workflows/build.yml
```diff
@@ -1343,6 +1343,9 @@ jobs:
     - name: Count the number of Patients with Stress Conditions at Inpatient Encounters using FHIR Search
       run: .github/scripts/count-patients.sh Condition "code=http://snomed.info/sct|73595000&encounter.class=http://terminology.hl7.org/CodeSystem/v3-ActCode|IMP" 13
 
+    - name: Download Condition Resources of common disorders from multiple patients
+      run: .github/scripts/download-resources-query.sh Condition "code=http://snomed.info/sct|444814009,http://snomed.info/sct|195662009,http://snomed.info/sct|10509002,http://snomed.info/sct|271737000,http://snomed.info/sct|40055000,http://snomed.info/sct|233604007,http://snomed.info/sct|389087006,http://snomed.info/sct|75498004,patient=DF6BKEGYCAVTJTJW,DF6BKIFPINSCUFMU,DF6BKG5XOZ6LIPCO,DF6BKEI33Y5DNPCL,DF6BKFIAZBTJQKBW,DF6BKLVM2F5ZMQ7E,DF6BKLBYOO56YBJP,DF6BKKPRTPMC56NT,DF6BKERT6ONPSDU6,DF6BKH6KO7KP3M4V,DF6BKE46NHZ5HJAD,DF6BKEUGGAXMEAVZ,DF6BKIPFOJVGBY6Z,DF6BKF6YOVUTEQOH,DF6BKFJWLYT25KWU" 2947
```
knoppiks: The patient IDs `DF6BKEGYCAVTJTJW` etc. look like magic here - where do they come from and are they stable?

knoppiks (changes requested):

.github/scripts/download-resources-query-sort.sh
alexanderkiel: done

.github/scripts/download-resources-query.sh
alexanderkiel: done

.github/workflows/build.yml
alexanderkiel: Right. They are not stable and the query contains an error that renders the patient search param useless. I uses a comma instead of an ampersand to separate the last SNOMED CT code and the patient search param code.

knoppiks (dismissed):

knoppiks (approved):

PR #2813, opened 2025-07-08
Mark Resources in Normal UI View as Subsetted

knoppiks:
Preview:
image image

### DIFF
--- a/modules/frontend-e2e/src/fhir.spec.ts
+++ b/modules/frontend-e2e/src/fhir.spec.ts
@@ -1,4 +1,4 @@
-import type {Page} from '@playwright/test';
+import type {Page, Locator} from '@playwright/test';
 import {expect, test} from '@playwright/test';
 
 test.beforeEach('Sign In', async ({page}) => {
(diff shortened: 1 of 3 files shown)


modules/frontend-e2e/src/fhir.spec.ts
```diff
@@ -31,6 +31,13 @@ async function expectInActiveNavLink(page: Page, name: string): Promise<void> {
     await expect(overviewLink).not.toHaveAttribute('aria-current', 'page');
 }
 
+async function expectBadge(locator: Locator): Promise<void> {
+    await expect(locator).toBeVisible();
+    await expect(locator).toHaveCSS('background-color', /rgb\(.*\)/);
+    await expect(locator).toHaveCSS('border-radius', '9999px');
```
alexanderkiel: If you use the better locator, you can remove the `toHaveRole` assertion.

modules/frontend/src/lib/history/subsetted-badge.svelte
```diff
@@ -0,0 +1,26 @@
...
+	function isSubsetted(entry: FhirObjectBundleEntry): boolean {
+		const tags = entry.resource?.meta?.tag;
+		return tags?.filter((c) => c.system === subsettedSystem)[0]?.code === 'SUBSETTED';
+	}
+
+	let { entry }: Props = $props();
+</script>
+
+{#if isSubsetted(entry)}
+	<p
```
alexanderkiel: We also need the `<div class="ml-2 flex flex-shrink-0">` from the other badges here. The Tailwind badge ha no margin, but the need the margin so that multiple badges don't touch.

modules/frontend-e2e/src/fhir.spec.ts
```diff
@@ -44,6 +51,7 @@ test('History Page', async ({page}) => {
 
     await expect(page).toHaveTitle("History - Blaze");
     await expect(page.getByText('Total:')).toBeVisible();
+    await expectBadge(page.getByText('subsetted').first());
```
alexanderkiel: Please use the better locator `page.getByRole('note').filter({hasText: 'subsetted'})`.

PR #2820, opened 2025-07-10
Configure Formatting for Typescript Modules

knoppiks:


alexanderkiel (changes requested):
Two things:

* the added dependencies in frontend-e2e/package.json are not inside the package-lock.json. please run a npm update before commit
* the docker-compose.yml entry in the frontend-e2e .prettierignore was added in the reformat commit but should be added in the first commit

PR #2929, opened 2025-08-21
Fix Next Link Detection

alexanderkiel:
In R5 the link relation in a bundle is a code instead of a string like in R4. Because a string without extensions is just a string itself, calling type/value isn't needed. But in R5 we need to call type/value because a code isn't a string.

Also fixed reflection warnings in test.

### DIFF
--- a/modules/fhir-client/test/blaze/fhir_client_test.clj
+++ b/modules/fhir-client/test/blaze/fhir_client_test.clj
@@ -1,5 +1,6 @@
 (ns blaze.fhir-client-test
   (:require
+   [blaze.async.flow :as flow]
    [blaze.fhir-client :as fhir-client]
    [blaze.fhir-client-spec]
    [blaze.fhir.parsing-context]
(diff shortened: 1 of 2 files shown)


modules/fhir-client/test/blaze/fhir_client_test.clj
knoppiks, line 1: Should there be a test, that shows that a code relation is now possible?

knoppiks (changes requested):

modules/fhir-client/test/blaze/fhir_client_test.clj
alexanderkiel: It's not relevant in R4, only in R5 / R6. As I wrote in the comment, in R4 the link relation is just a string. Because strings without extensions are already normal strings, calling `type/value` on them doesn't do anything. In R5 however, it's needed, because a code isn't a string.

knoppiks (approved):

PR #2983, opened 2025-09-25
Validation Support for Resources

### LINKED ISSUE #968: Support Profile Validation
The user should be able to upload profiles that would be used to validate resources on ingest.

xmyno:
This PR adds validation support for resources based on StructureDefinitions.

Closes #968


alexanderkiel:
Done by #3697 because we don't like to depend on HAPI internally.

PR #2990, opened 2025-09-30
Update Frontend Dependencies

alexanderkiel:
The main work was to satisfy the new linter rule to use resolve. However, I did not manage to get this working in every place. In later updates we might have to do more here.


EmteZogaf (approved):
lgtm

Issue #2704, opened 2025-05-29
Replace Integrant Prep with Expand


alexanderkiel:
`integrant.core/prep` is deprecated in favour of `expand`. We use `prep` in `blaze.system/init!`.

Issue #3022, opened 2025-10-07
Explore the Usage of new Integrant Features


alexanderkiel:
Currently, we only use `init` from Integrant. However, there are profiles, expand and vars. We should think about using that instead of our home-grown feature toggle system.

PR #3023, opened 2025-10-07
Migrate Integrant to v1.0

alexanderkiel:
We have to remove the use of `prep` which we actually never used anyway. So the migration is simpler as expected. In #3022 we can explore the use of the new Integrant features.

Closes: #2704


EmteZogaf (approved):
lgtm

Issue #3090, opened 2025-11-06
Authentication: check more than the first key


gongham:
Instead of only checking the first key, can you loop through the keys and authenticate if one of them work?
And return error if all the keys fail?

I am happy to fork, I am just struggling with interpreting clojure. Any pointers would help.

Thank you

gongham:
Here is the code snippet to replace public-key in impl.clj:

```clojure
(defn- public-key
  "Try all signing JWKs from JWKS until one successfully converts into a PublicKey.
  Returns the first successful PublicKey, or an anomaly if all fail."
  [[jwks-uri jwks-document]] 

  (let [sig-jwks (filter #(= "sig" (:use %)) (:keys jwks-document))]
    (if (seq sig-jwks)
      (loop [[jwk & remaining] sig-jwks]
        (when jwk
          (try
            ;; Attempt conversion
            (keys/jwk->public-key jwk)
            (catch Exception e
              (log/warn "Failed to convert JWK with kid:" (:kid jwk)
                        "- error:" (ex-message e))
              ;; move to next JWK
              (recur remaining)))))
      (ba/fault (missing-key-msg jwks-uri)))))
```

PR #3094, opened 2025-11-07
Authentication: check more than the first key

alexanderkiel:
Closes: #3090


gongham:
Thanks for the swift response @alexanderkiel 
When you have a chance, could you please create a release for this?
We are trying to deploy with authentication.
Thanks so much!

alexanderkiel:
> Thanks for the swift response @alexanderkiel When you have a chance, could you please create a release for this? We are trying to deploy with authentication. Thanks so much!

I'll release the version 1.2.0 in the next two weeks. You can always use the current container on main via `sample/blaze:main`.

PR #3097, opened 2025-11-09
Prevent Cache Thrashing During Large Resource Scans

alexanderkiel:


modules/db/test/blaze/db/resource_cache_test.clj
```diff
@@ -129,6 +132,98 @@
...
+
+        ;; insert every second patient
+        (doseq [hash (keys patients)
+                :when (< (rand) 0.5)]
+          @(rc/get cache [:fhir/Patient hash :complete]))
+
+        (given @(st/with-instrument-disabled
+                  (rc/multi-get-skip-cache-insertion
+                   cache (mapv #(vector :fhir/Patient % :complete) (keys patients))))
+          count := 100)))))
```
knoppiks: Should we check, that the cache only contains cached resources?

alexanderkiel: done

Issue #3086, opened 2025-11-04
Prevent Cache Thrashing During Large Resource Scans


alexanderkiel:
### Problem

The resource cache experiences thrashing when scan operations access more resources than the cache can hold. Since the cache typically stores only ~1% of total resources, any scan exceeding this threshold causes continuous eviction and replacement of cache entries, degrading performance.

**Current behavior:**
- All resource access goes through the cache, causing immediate insertion on load
- FHIR Search paging can retrieve large result sets across multiple pages
- Each page load populates the cache, displacing potentially useful entries
- Cache hit ratio degrades as useful entries are evicted by scan operations

### Proposed Solution

Introduce a cache-aware bypass path for resource access that looks up resources in the cache but doesn't insert misses.

**Implementation strategy:**
- **First page results**: Continue using normal cached access
  - Limited result size (per page size limits)
  - Higher likelihood of repeated access
  - Maintains cache effectiveness for common queries
  
- **Subsequent pages**: Use bypass path
  - Checks cache first and returns cached resources if present
  - On cache miss, retrieves resource directly without cache insertion
  - Prevents displacement of frequently-accessed resources
  - Maintains cache stability during large result set traversal

### Success Criteria

- Cache hit ratio remains stable during multi-page FHIR Search operations
- First page results remain cached for potential reuse
- Large scans no longer evict useful cache entries
- Cached resources are still utilized when available

PR #3098, opened 2025-11-10
Refactor d/pull-many

alexanderkiel:
It now requires a concrete collection instead of accepting a reducible one in order to not having to first realize it.

The variant argument is now an option map in preparation for later additions from #3086.

### DIFF
--- a/modules/db/src/blaze/db/api.clj
+++ b/modules/db/src/blaze/db/api.clj
@@ -528,17 +528,20 @@
   "Returns a CompletableFuture that will complete with a vector of all resources
   of all `resource-handles` in the same order.
 
-  Optional, `variant` can be given which is either a content variant like
-  :complete or :summary or a list of top-level keys to return instead of all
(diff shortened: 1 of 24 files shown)


modules/db/src/blaze/db/api.clj
```diff
@@ -528,17 +528,20 @@
...
-  Optional, `variant` can be given which is either a content variant like
-  :complete or :summary or a list of top-level keys to return instead of all
-  keys (elements). Certain mandatory and modifier elements are returned
-  regardless of if they are specified in `variant`. In addition the resources
-  are marked with the tag SUBSETTED in this case.
+  The following options are available:
+  * :variant - which is either :complete or :summary
+  * :elements - a list of top-level keys to return instead of all
+
+  Certain mandatory and modifier elements are returned regardless of if they are
```
knoppiks: `of if` does not sound right to me.
```suggestion
  Certain mandatory and modifier elements are returned regardless of not being
```

PR #3156, opened 2025-12-01
Dynamically Lower Page Size on Too Many Includes

knoppiks:


modules/interaction/test/blaze/interaction/search_type_test.clj
```diff
@@ -2696,6 +2698,109 @@
             [:issue 0 :code] := #fhir/code "too-costly"
             [:issue 0 :diagnostics] := #fhir/string "Inclusion(s) would return more than 10,000 resources which is too costly to output. Please either lower the page size or use $graphql or $graph operations."))))
 
+    (testing "other include error"
+      (with-redefs [search-include/add-includes
+                    (constantly (ba/conflict "Some problem 133007"))]
```
alexanderkiel: please use "msg-133007" it's used very often this way

modules/interaction/test/blaze/interaction/search_type_test.clj
```diff
@@ -2696,6 +2698,109 @@
...
+                              "_count" "50" "__t" "3"
+                              "__page-id" "1"})
+                   (link-url body "next"))))
+
+          (testing "second page"
+            (let [{:keys [status body]}
+                  @(handler
+                    {::reitit/match (match-of "Patient")
+                     :params
+                     {"_revinclude" "Observation:subject" "_count" "50"
```
alexanderkiel: if it works this way, that would be a bug. It should be necessary to use this to get to the next page:

```clojure
{::reitit/match patient-page-match
                 :path-params (page-path-params page-id-cipher {"_revinclude" "Observation:subject" "_count" "50" "__t" "3" "__page-id" "1"})}
```

modules/interaction/test/blaze/interaction/search_type_test.clj
```diff
@@ -2696,6 +2698,110 @@
...
+            (is (= 5001 (count (:entry body)))))
+
+          (testing "the first entry is the first matched Patient"
+            (given (-> body :entry first)
+              [:fullUrl :value] := (str base-url context-path "/Patient/0")
+              [:resource :fhir/type] := :fhir/Patient
+              [:search :mode] := #fhir/code "match"))
+
+          (testing "has a self link with reduced page-size"
+            (is (= (str base-url context-path "/Patient?_revinclude=Observation%3Asubject&_count=1")
```
alexanderkiel: after thinking about it again, I would say, we should keep the original page count. Requesting a certain count doesn't mean you get it. This behaviour is even preferred in the spec:

> Servers SHALL NOT return more resources than requested, even if they don't support paging, but may return less than the client requested. The server should repeat the original _count parameter in its returned page links so that subsequent paging requests honor the original _count.

.github/scripts/search-include-too-costly.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -e
+
```
alexanderkiel: please describe here what the script does

.github/scripts/search-include-too-costly.sh
```diff
@@ -0,0 +1,21 @@
...
+script_dir="$(dirname "$(readlink -f "$0")")"
+. "$script_dir/util.sh"
+
+base="http://localhost:8080/fhir"
+query="${1//[[:space:]]/}"
+result="$(curl -sSf "$base/Patient?$query&_count=2&_revinclude=Observation:subject")"
+
+patient_count=$(echo "$result" | jq -r '[.entry[].resource | select(.resourceType == "Patient")] | length')
```
alexanderkiel: please look at sreach mode match instead

Issue #2434, opened 2025-03-03
Consider RocksDB Config for Large Databases


alexanderkiel:
Currently the SST target file size for most column families is set to the default of 64 MiB. However that size will result in about 65k files for database sizes of about 2 TiB. 

It might be a good idea to offer an alternative set of RocksDB configs for large databases. That set could multiply all sizes by a factor of 4.

PR #3191, opened 2025-12-13
Introduce Scale Factor Config for Large Databases

alexanderkiel:
Closes: #2434


docs/tuning-guide.md
knoppiks, line 1: We recently talked about renaming this section of the docs. "Tuning" sounds like an optional step. But scaling the configuration to the number of patients/resources intended to be stored is fundamental for a stable production deployment.

Issue #3242, opened 2026-01-14
Update Java to v25


alexanderkiel:
Migrate health checks to `wget` which will be available in the frontend and the backend container and has a smaller footprint as `curl`. The command `curl` should be still available for backwards compatibility and will be removed in the next minor release. We can do that because `curl` was never advertised as being available.

PR #3243, opened 2026-01-14
Update Java to v25

alexanderkiel:
Closes: #3242


.github/integration-test-kds/docker-compose.yml
knoppiks, line 1: Why remove the health checks?

alexanderkiel: Because I like to do work in between starting. So I can't use `--wait` and so don't need health checks.

knoppiks (dismissed):

Issue #3287, opened 2026-01-26
Improve Error Handling at Patient Type Queries


alexanderkiel:

alexanderkiel:
Will be solved in #3294.

PR #3288, opened 2026-01-26
Improve Error Handling at Patient Type Queries

alexanderkiel:
Because the function execute-query can return an anomaly now, I had to disable it on non-batch databases. It's just not possible to handle batch database opening and closing if a function can return a reducible collection and an anomaly at the same time.

I refactored code using execute-query to ensure a batch database is always used. That should even improve performance because we didn't use a batch databases at search interactions.

Closes: #3287


alexanderkiel:
Will be solved in #3329.

Issue #3291, opened 2026-01-28
Fix FHIR Search Optimisation Problem


alexanderkiel:
FHIR Search queries that include criteria with patient references, are optimised to use patient compartment indices. In case only the patient reference criteria or no other token criteria is given, the CompartmentResourceType index is used. However this index has the problem that the hash prefix (version of the resource) is missing. Because of this missing version information, resources are found even if the current version contains no longer the patient reference. At the end, we can not use the CompartmentResourceType index.

The goal is to not use the the CompartmentResourceType index while optimising FHIR search queries.

PR #3297, opened 2026-01-28
Fix FHIR Search Optimisation Problem

alexanderkiel:
Closes: #3291


FloSeidel:
This seems to fix the issue we had :)

PR #3390, opened 2026-02-25
Add support for CQL In operator

paulolaup:


paulolaup:
Favored solution is proper normalization before expression compilation

alexanderkiel:
Solved by #3391

Issue #3397, opened 2026-03-03
Add Error Handling Docs to FHIR Search API


alexanderkiel:
The documentation at [FHIR API – Interaction – Search Type](https://samply.github.io/blaze/api/interaction/search-type.html) is missing a [Handling Errors](https://hl7.org/fhir/R4/search.html#errors) section like the FHIR Specification.

Please add such a section with a concise description how the Prefer header should be set to get error messages on unknown/unsupported search parameters.

PR #3398, opened 2026-03-03
Add Error Handling Docs to FHIR Search API

alexanderkiel:
Closes: #3397

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -44,6 +44,8 @@ Rigorous adherence to these patterns is required:
 
 ## Verification & Workflow
 
+When starting to work on an issue, you can use the GitHub CLI to fetch the issue details: `gh issue view <issue-number>`
+
 Before finishing a task, ensure the following commands pass:
 
 1.  **Format:** `make fmt`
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -44,13 +44,18 @@ Rigorous adherence to these patterns is required:
...
 1.  **Format:** `make fmt`
 2.  **Lint:** `make lint` (Uses `clj-kondo`)
 3.  **Test:** `make test` (Runs module and root tests)
 4.  **Coverage:** `make test-coverage` (Checks for adequate test coverage)
 
-## General Rules
+After verification, when working on an issue:
 
-* **Do not commit changes:** Never commit changes to the git repository. The user will handle committing.
+1. Create a feature branch using the GitHub CLI: `gh issue develop <issue-number> --checkout`
```
trobanga, line 58: Here you could think about using [git worktrees](https://git-scm.com/docs/git-worktree)

AGENTS.md
```diff
@@ -44,13 +44,18 @@ Rigorous adherence to these patterns is required:
...
+2. Commit the changes: `git add .` and `git commit`
```
trobanga, line 59: We should make sure that the AI only creates exactly one commit per issue and amends in case of changes

AGENTS.md
alexanderkiel: Yes, I would do that as next step.

AGENTS.md
alexanderkiel: Added that.

PR #3523, opened 2026-04-01
Periodically Log JVM Memory and GC Usage

### LINKED ISSUE #3522: Periodically Log JVM Memory and GC Usage
## Background

The health of every JVM-based service — and so also Blaze — depends heavily on heap memory usage. When the heap becomes too full, the JVM triggers long GC pauses during which Blaze appears unresponsive from the outside. Diagnosing such incidents without memory usage history is difficult and leads to a frustrating experience for Blaze operators.

## Current State

Blaze already exports JVM memory and GC metrics via the Prometheus endpoint (port 8081) through the `metrics` module, which registers `MemoryPoolsExports` and `GarbageCollectorExports` from the Prometheus Java client. However, not every operator sets up Prometheus monitoring, and there is currently no fallback for surfacing this critical information in the application log.

## Proposed Solution

Periodically log heap and non-heap memory usage as well as GC activity using `java.lang.management.ManagementFactory` (the same JVM MXBeans that back the Prometheus exports). The existing `blaze.scheduler` module provides `schedule-at-fixed-rate` and is the natural mechanism for this.

### Logging Schedule

| Condition | Level | Interval |
|---|---|---|
| Normal operation (heap usage < 80%) | `DEBUG` | every 5 minutes |
| Elevated usage (heap usage ≥ 80%) | `WARN` | every 1 minute |

### Log Content (per interval)

- Heap: used / committed / max (absolute values and percentage of max)
- Non-heap: used / committed
- GC: collection count and total pause time per collector (e.g. G1 Young Generation, G1 Old Generation)

### Example Log Lines

```
DEBUG blaze.jvm-metrics - Heap: 1.2 GB used / 2.0 GB committed / 4.0 GB max (30%), Non-Heap: 120 MB used / 128 MB committed, GC: G1 Young 42 collections 1.2s, G1 Old 1 collection 0.3s
WARN  blaze.jvm-metrics - High heap usage - Heap: 3.7 GB used / 4.0 GB committed / 4.0 GB max (93%), Non-Heap: 125 MB used / 132 MB committed, GC: G1 Young 310 collections 18.4s, G1 Old 12 collections 4.1s
```

## Implementation Sketch

A new `jvm-metrics-logger` module that:

1. Reads `MemoryMXBean`, `MemoryPoolMXBean`s, and `GarbageCollectorMXBean`s on each tick.
2. Computes heap usage percentage and selects the log level / interval accordingly.
3. Is wired into the Integrant system and receives the `blaze/scheduler` dependency.

## Acceptance Criteria

- [ ] At least every 5 minutes a `DEBUG` log line is emitted with heap, non-heap, and GC summary.
- [ ] When heap usage is ≥ 80% the interval drops to 1 minute and the level becomes `WARN`.
- [ ] The log output is human-readable and includes absolute values plus the percentage of max heap.
- [ ] The feature is covered by unit tests (mock MXBeans or deterministic scheduler).
- [ ] The logging interval and warning threshold are configurable via environment variables with sensible defaults.

## Integration Tests

- test that at least one such metrics line is logged at the end of the integration-test job

alexanderkiel:
Closes: #3522


modules/jvm-metrics-logger/src/blaze/jvm_metrics_logger.clj
```diff
@@ -0,0 +1,100 @@
...
+  (let [heap ^MemoryUsage (.getHeapMemoryUsage (ManagementFactory/getMemoryMXBean))
+        pct  (heap-pct heap)]
+    (if (>= pct warn-threshold)
+      (log/warn "High heap usage -" (heap-usage-string))
+      (log/info (heap-usage-string))))
+  nil)
+
+(defmethod m/pre-init-spec :blaze/jvm-metrics-logger [_]
+  (s/keys :req-un [:blaze/scheduler]
+          :opt-un [::normal-interval ::warn-interval ::warn-threshold]))
```
knoppiks:
```suggestion
          :opt-un [::default-interval ::warn-interval ::warn-threshold]))
```

modules/jvm-metrics-logger/src/blaze/jvm_metrics_logger.clj
```diff
@@ -0,0 +1,100 @@
...
+  [_ {:keys [scheduler default-interval warn-interval warn-threshold]
+      :or {default-interval (time/minutes 5)
+           warn-interval (time/minutes 1)
+           warn-threshold 80}}]
+  (log/info "Start JVM metrics logger")
+  (let [default-nanos  (time/as default-interval :nanos)
+        last-info-log (atom 0)]
+    {:future
+     (sched/schedule-at-fixed-rate
+      scheduler
```
trobanga, line 36: The scheduled function reads `ManagementFactory/getMemoryMXBean` to compute `pct`, then `log-metrics\!` reads it again and recomputes `pct` independently. Between the two reads, heap usage could cross the threshold boundary — e.g., the scheduler decides "above threshold → WARN", but by the time `log-metrics\!` reads, usage dropped and it logs at INFO instead.

Consider reading the `MemoryUsage` once and threading it through, or refactoring so the level decision and the log message use the same snapshot.

modules/jvm-metrics-logger/test/blaze/jvm_metrics_logger_test.clj
```diff
@@ -0,0 +1,100 @@
...
+      (is (re-find #"Non-Heap:" result))
+      (is (re-find #"\d+%" result))))
+
+  (testing "includes GC info when collectors are present"
+    (let [result (jml/heap-usage-string)]
+      (is (re-find #"GC:" result)))))
+
+(deftest log-metrics-test
+  (testing "logs at info level when below threshold"
+    (is (nil? (jml/log-metrics! 100))))
```
trobanga: The `log-metrics-test` only asserts the return value is `nil`, and the `scheduled-*-task-test` tests assert `(is true)` after a sleep. None of these verify that:
- The correct log level (INFO vs WARN) was used
- Logging actually occurred
- The rate-limiting logic throttles info logs to `default-interval`

Since this feature is specifically about logging at the right level under the right conditions, capturing log output (e.g., via a Timbre appender or `with-redefs`) would make these tests much more meaningful.

modules/jvm-metrics-logger/src/blaze/jvm_metrics_logger.clj
```diff
@@ -0,0 +1,100 @@
...
+      (fn []
+        (let [heap ^MemoryUsage (.getHeapMemoryUsage (ManagementFactory/getMemoryMXBean))
+              pct  (heap-pct heap)
+              now  (System/nanoTime)]
+          (cond
```
trobanga: Nit: `last-info-log` is initialized to `0`, relying on `(- (System/nanoTime) 0)` being large enough to trigger the first log. `System/nanoTime` has no guaranteed epoch — its contract only says *differences* between two calls are meaningful. Initializing to `(- (System/nanoTime) default-nanos)` would make the "log immediately on first tick" intent explicit and correct by contract.

modules/jvm-metrics-logger/src/blaze/jvm_metrics_logger.clj
```diff
@@ -0,0 +1,100 @@
...
+            (if (= 1 cnt) "collection" "collections") time-s)))
+
+(defn heap-usage-string
+  "Returns a human-readable string summarising the current JVM heap, non-heap
+  and GC metrics. Exported as a public function to facilitate testing."
+  []
+  (let [heap     ^MemoryUsage (.getHeapMemoryUsage (ManagementFactory/getMemoryMXBean))
+        non-heap ^MemoryUsage (.getNonHeapMemoryUsage (ManagementFactory/getMemoryMXBean))
+        pct      (heap-pct heap)
+        gcs      (seq (ManagementFactory/getGarbageCollectorMXBeans))
```
trobanga: Nit: `ManagementFactory/getMemoryMXBean` is called twice here (returns the same singleton). Binding it once would be slightly cleaner:
```clojure
(let [mbean    (ManagementFactory/getMemoryMXBean)
      heap     ^MemoryUsage (.getHeapMemoryUsage mbean)
      non-heap ^MemoryUsage (.getNonHeapMemoryUsage mbean)
```

resources/blaze.edn
```diff
@@ -463,6 +463,16 @@
...
+  :blaze/jvm-metrics-logger
+  {:scheduler #blaze/ref :blaze/scheduler
+   :default-interval #blaze/cfg ["JVM_METRICS_LOGGER_DEFAULT_INTERVAL"
+                                java-time.api/duration?
+                                "PT5M"]
+   :warn-interval #blaze/cfg ["JVM_METRICS_LOGGER_WARN_INTERVAL"
+                              java-time.api/duration?
+                              "PT1M"]
+   :warn-threshold #blaze/cfg ["JVM_METRICS_LOGGER_WARN_THRESHOLD" pos-int? 80]}
+
```
trobanga, line 472: Nit: The coercer here is `pos-int?` which accepts any positive integer, but the spec constrains to 1–99. Setting `JVM_METRICS_LOGGER_WARN_THRESHOLD=100` would pass coercion but fail with a spec error at startup. Not blocking (the spec catches it), but the resulting error message won't be as friendly as a dedicated bounds check.

modules/jvm-metrics-logger/src/blaze/jvm_metrics_logger.clj
alexanderkiel: done

modules/jvm-metrics-logger/test/blaze/jvm_metrics_logger_test.clj
alexanderkiel: done

modules/jvm-metrics-logger/src/blaze/jvm_metrics_logger.clj
alexanderkiel: done via a combination of interval and warn-factor

modules/jvm-metrics-logger/src/blaze/jvm_metrics_logger.clj
alexanderkiel: done

resources/blaze.edn
alexanderkiel: done

trobanga (dismissed):

trobanga (approved):

PR #3560, opened 2026-04-13
Cassandra Read Consistency ONE Causes Retries Under Load

### LINKED ISSUE #3555: Cassandra Read Consistency ONE Causes Retries Under Load
## Problem

The GET and PUT statements in both the Cassandra resource store and page store use mismatched consistency levels that do **not** guarantee strong consistency. Under high load, this causes massive retry storms that degrade performance and can lead to request timeouts.

## Root Cause

With `replication_factor = 3`, the current consistency levels are:

| Operation | Consistency | Code location |
|-----------|-------------|---------------|
| PUT (write) | `TWO` | `modules/db-resource-store-cassandra/src/blaze/db/resource_store/cassandra/statement.clj` |
| GET (read) | `ONE` | same file |

The same pattern exists in `modules/page-store-cassandra/src/blaze/page_store/cassandra/statement.clj`.

This does **not** satisfy Cassandra's strong consistency requirement:

```
read_nodes + write_nodes > RF  →  1 + 2 = 3 ≯ 3
```

The reading node may be the one that hasn't received the write yet. The code acknowledges this in a comment:

> *"The consistency level is set to ONE because reads are retried if nothing was found. Because rows are never updated, strong consistency isn't needed."*

Under light load, the 5-retry exponential backoff (100ms → 200ms → 400ms → 800ms → 1600ms) masks the problem. But under high write throughput — such as concurrent bundle imports or any heavy ingestion workload — Cassandra's async replication lags significantly, causing nearly every GET to miss and triggering the full retry sequence (~3.1s per resource lookup).

## Evidence

Observed in CI (`integration-test (distributed, cql)`):
- ~**103,000** `resource-store-cassandra-get` retry warnings during an ~8-minute test run
- Nearly every Cassandra GET went through multiple retries
- Max request latency hit **2m0.039s**, exceeding the 120s proxy timeout and producing a **504**

This is not specific to the distributed test setup — any deployment with `replication_factor ≥ 2` and sufficient write load will exhibit the same retry storms and latency degradation.

## Proposed Fix

Change the GET statement from `ConsistencyLevel/ONE` to `ConsistencyLevel/TWO` in both:
- `modules/db-resource-store-cassandra/src/blaze/db/resource_store/cassandra/statement.clj`
- `modules/page-store-cassandra/src/blaze/page_store/cassandra/statement.clj`

```
read=TWO, write=TWO, RF=3  →  2 + 2 = 4 > 3  ✓
```

With read=TWO and write=TWO, any 2 nodes queried on a read must intersect with the 2 nodes written to, guaranteeing the data is always found. This eliminates the false not-found retries entirely. The retry mechanism can remain as a safety net for genuine transient errors (e.g. `::anom/busy`).

alexanderkiel:
Closes: #3555

### DIFF
--- a/modules/cassandra/src/blaze/cassandra.clj
+++ b/modules/cassandra/src/blaze/cassandra.clj
@@ -4,13 +4,12 @@
    [blaze.anomaly :as ba]
    [blaze.cassandra.session :as session]
    [blaze.util :refer [str]]
-   [clojure.string :as str]
-   [cognitect.anomalies :as anom])
+   [clojure.string :as str])
   (:import
    [com.datastax.oss.driver.api.core
     CqlSession DriverTimeoutException RequestThrottlingException]
    [com.datastax.oss.driver.api.core.cql
-    AsyncResultSet PreparedStatement Row Statement SimpleStatement]
+    AsyncResultSet PreparedStatement Row SimpleStatement Statement]
    [com.datastax.oss.driver.api.core.servererrors WriteTimeoutException]))
 
 (set! *warn-on-reflection* true)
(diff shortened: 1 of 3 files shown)


alexanderkiel:
done by #3563

PR #3577, opened 2026-04-16
Implement filter parameter on ValueSet/$expand

### LINKED ISSUE #3570: Implement filter parameter on ValueSet/$expand
## Summary

Implement the `filter` parameter on `ValueSet/$expand` ([spec](https://hl7.org/fhir/6.0.0-ballot4/valueset-operation-expand.html)) to enable typeahead/autocomplete search over concept displays and designations in UIs.

## Requirements

- The `filter` parameter accepts a free-text string and restricts the returned concepts to those matching the filter
- Matching should be intuitive for humans in a UI (word-prefix matching: typing "blood pres" matches "Systolic blood pressure")
- Results should be ranked by relevance (exact/shorter matches first)
- Search must run in-process on the JVM — no external services (e.g. ElasticSearch)
- The search index should be pre-built from code system concepts (display + designation values) and cached

## Approach

Use **Apache Lucene** with in-memory `ByteBuffersDirectory`:

- Add `org.apache.lucene/lucene-core` and `org.apache.lucene/lucene-analysis-common` (~4MB) to `modules/terminology-service-local/deps.edn`
- Create a new namespace `blaze.terminology-service.local.search-index` that builds a Lucene index from a code system's concepts and queries it with the filter string
- Integrate into the expand pipeline in `blaze.terminology-service.local.value-set.expand` — after expanding concepts, filter them through the search index
- Cache the Lucene index alongside the graph (keyed by `[code-system-id, version]`)

Lucene can later be switched from `ByteBuffersDirectory` (in-memory) to `MMapDirectory` (disk-cached) if needed, with no API changes. A protocol-based abstraction could also allow swapping in ElasticSearch later.

## Module

`modules/terminology-service-local`

alexanderkiel:
Closes: #3570

### DIFF
--- a/modules/frontend-e2e/docker-compose.yml
+++ b/modules/frontend-e2e/docker-compose.yml
@@ -49,6 +49,8 @@ services:
       ENABLE_ADMIN_API: "true"
       ENABLE_TERMINOLOGY_SERVICE: "true"
       ENABLE_TERMINOLOGY_LOINC: "true"
+      ENABLE_TERMINOLOGY_SNOMED_CT: "true"
+      SNOMED_CT_RELEASE_PATH: "/app/sct-release"
       OPENID_PROVIDER_URL: "https://keycloak.localhost/realms/blaze"
       OPENID_CLIENT_TRUST_STORE: "/app/keycloak-trust-store.p12"
       OPENID_CLIENT_TRUST_STORE_PASS: "insecure"
@@ -56,6 +58,7 @@ services:
     - "keycloak.localhost:host-gateway"
     volumes:
     - "blaze-data:/app/data"
+    - "./SnomedCT_InternationalRF2_PRODUCTION_20250701T120000Z:/app/sct-release"
     - "../ingress/keycloak-trust-store.p12:/app/keycloak-trust-store.p12:ro"
     healthcheck:
       test: [ "CMD", "wget", "--spider", "http://localhost:8080/health" ]
@@ -66,12 +69,18 @@ services:
 
   keycloak:
     image: "quay.io/keycloak/keycloak:26.4.2@sha256:3617b09bb4b7510a8d8d9b9fc5707399e2d70688dbcc2f8fb013a144829be1b9"
-    command: "start --import-realm --hostname=keycloak.localhost --proxy-headers forwarded --http-enabled true"
+    command: "start --import-realm --hostname=keycloak.localhost --proxy-headers forwarded --http-enabled true --health-enabled=true"
     environment:
       KEYCLOAK_ADMIN: admin
       KEYCLOAK_ADMIN_PASSWORD: admin
     volumes:
     - "./keycloak/import:/opt/keycloak/data/import"
+    healthcheck:
+      test: [ "CMD-SHELL", "exec 3<>/dev/tcp/127.0.0.1/9000; echo -e 'GET /health/ready HTTP/1.1\r\nHost: localhost:9000\r\nConnection: close\r\n\r\n' >&3; grep -q '\"status\": \"UP\"' <&3" ]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+      start_period: 30s
 
 volumes:
   blaze-data:
--- a/modules/frontend-e2e/src/value-set.spec.ts
+++ b/modules/frontend-e2e/src/value-set.spec.ts
@@ -1,0 +1,218 @@
+import { expect, type Locator, type Page, test } from '@playwright/test';
+
+function breadcrumbItem(page: Page, text: string): Locator {
+  return page.getByLabel('Breadcrumb').getByRole('listitem').filter({ hasText: text });
+}
+
+test.beforeEach('Sign In', async ({ page }) => {
+  await page.goto('/fhir/ValueSet');
+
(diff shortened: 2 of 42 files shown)


modules/frontend-e2e/src/value-set.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+        page,
+        'MII VS Labor Interpretationsbeeinflussende Eigenschaften SNOMEDCT v2026.0.0'
+      )
+    ).toBeVisible();
+    await page.getByRole('button', { name: 'Operations' }).click();
+    await page.getByRole('menuitem', { name: '$expand' }).click();
+
+    await page.getByLabel('Filter').fill('lipemic');
+    await page.getByRole('button', { name: 'Submit' }).click();
+
```
knoppiks, line 210: Add a check, that the filter still contains the search term after submit. 
When testing this I discovered that the filter is momentarily empty, while the search results are loaded.

modules/frontend-e2e/docker-compose.yml
```diff
@@ -66,12 +69,18 @@ services:
...
   keycloak:
     image: "quay.io/keycloak/keycloak:26.4.2@sha256:3617b09bb4b7510a8d8d9b9fc5707399e2d70688dbcc2f8fb013a144829be1b9"
-    command: "start --import-realm --hostname=keycloak.localhost --proxy-headers forwarded --http-enabled true"
+    command: "start --import-realm --hostname=keycloak.localhost --proxy-headers forwarded --http-enabled true --health-enabled=true"
     environment:
       KEYCLOAK_ADMIN: admin
       KEYCLOAK_ADMIN_PASSWORD: admin
     volumes:
     - "./keycloak/import:/opt/keycloak/data/import"
+    healthcheck:
```
knoppiks, line 78: Adding the healthcheck seems out-of-scope for the issue.

modules/frontend/src/routes/ValueSet/$expand/+page.server.ts
```diff
@@ -0,0 +1,135 @@
...
+      url,
+      valueSetVersion,
+      filter,
+      property,
+      displayLanguage,
+      systemVersion,
+      includeDesignations,
+      includeDefinition,
+      activeOnly,
+      excludeNested,
```
knoppiks: Duplicate code. Can we somehow prepare the map and inject `incorrect`+`msg` on failure and `valueSet` on success?

modules/frontend/src/routes/ValueSet/$expand/+page.svelte
```diff
@@ -0,0 +1,99 @@
...
+              •
+              {#if contains.version}
+                <p>{contains.version}</p>
+                •
+              {/if}
+              <p>{contains.code}</p>
+            </div>
+            {#if contains.designation}
+              <ul>
+                {#each contains.designation as designation (designation.language + '-' + designation.use + '-' + designation.value)}
```
knoppiks, line 89: We should use keys in each blocks: https://svelte.dev/docs/svelte/best-practices#Each-blocks

modules/terminology-service-local/src/blaze/terminology_service/local/code_system/sct.clj
```diff
@@ -30,65 +31,44 @@
...
+
+(defn- find-concept
+  [{{:keys [module-dependency-index concept-index]} :sct/context
+    :sct/keys [module-id version]}]
+  (partial context/find-concept module-dependency-index concept-index module-id
+           version))
+
+(defn- find-acceptability
+  [{{:keys [acceptability-index]} :sct/context :sct/keys [version]}]
+  (partial context/find-acceptability acceptability-index version))
```
knoppiks, line 54: do we really need `partial` here?

knoppiks (changes requested):

modules/frontend-e2e/src/value-set.spec.ts
alexanderkiel: I don't think the check will do anything, because the test will just wait until it's available and not detect the flickering. Nevertheless solved the flickering issue.

modules/frontend-e2e/docker-compose.yml
alexanderkiel: yes, but I find it useful

modules/frontend/src/routes/ValueSet/$expand/+page.server.ts
alexanderkiel: done

modules/frontend/src/routes/ValueSet/$expand/+page.svelte
alexanderkiel: already done

knoppiks: nevermind, the key is already given in parens

modules/terminology-service-local/src/blaze/terminology_service/local/code_system/sct.clj
alexanderkiel: partial is good, safes one class to load

knoppiks (approved):

PR #3747, opened 2026-06-24
Add `parameters` Input to $evaluate-measure

### LINKED ISSUE #3745: Add `parameters` Input to $evaluate-measure
## Background

The FHIR R5 `Measure/$evaluate-measure` operation adds an input parameter named [`parameters`][1] that lets a client pass values into the CQL evaluation. This allows a single measure to be evaluated with different inputs (e.g. thresholds, codes, date boundaries) without editing and re-publishing the measure's library.

Blaze is an R4 server and the [R4 `$evaluate-measure`][2] definition does not include `parameters`, but the parameter is a useful, forward-compatible addition. This issue tracks adding support for it.

### R5 parameter definition

| Property | Value |
| --- | --- |
| Name | `parameters` |
| Use | `in` |
| Cardinality | `0..1` |
| Type | `Parameters` |

> Any input parameters for the evaluation. Parameters defined in this input will be made available by name to the CQL expression. Parameter types are mapped to CQL as specified in the Using CQL topic of the Clinical Reasoning Module. If a parameter appears more than once in the input Parameters resource, it is represented with a List in the input CQL. If a parameter has parts, it is represented as a Tuple in the input CQL.

## Current State

Today a measure's CQL parameters can only take the **default values** declared in the library — there is no way to override them per request.

- **Compilation** — `blaze.elm.compiler.library/compile-library` returns `:parameter-default-values` (a name → compiled-default map) extracted from the library's `parameters.def` (`modules/cql/src/blaze/elm/compiler/library.clj`).
- **Context build** — `blaze.fhir.operation.evaluate-measure.measure/enhance-context` sets `:parameters` to those defaults and resolves them into the expressions before optimization:
  ```clojure
  (assoc context ... :parameters parameter-default-values ...)
  ;; ...
  (-> (remove-unused-defs expression-defs measure)
      (library/resolve-params parameter-default-values)
      (optimize))
  ```
  (`modules/operation-measure-evaluate-measure/src/blaze/fhir/operation/evaluate_measure/measure.clj`)
- **Param parsing** — `blaze.fhir.operation.evaluate-measure.middleware.params` parses only `periodStart`, `periodEnd`, `measure`, `reportType` and `subject`, producing the `:blaze.fhir.operation.evaluate-measure/params` map. The `parameters` resource is currently ignored, and `get-param-value-from-resource` only handles primitive-valued (`parameter.value[x]`) entries via `(-> % :value :value)`, not resource-valued (`parameter.resource`) entries.

So the request-side `params` and the CQL-side `:parameters` are disconnected: there is no path for a client-supplied value to reach `resolve-params`.

## Proposed Change

Thread client-supplied parameters from the request through to CQL evaluation, overriding the library defaults by name.

The `parameters` input is **only supported via `POST`** with a `Parameters` body — it is a resource-typed input nested in `parameter.resource` and cannot be expressed in the `GET` query string.

### Scope

- [ ] **Parse `parameters` in the middleware** (`middleware/params.clj`): read the nested `Parameters` resource from `parameter.resource` of the POSTed `Parameters` body. Add it to the `:blaze.fhir.operation.evaluate-measure/params` map (e.g. under `:parameters`).
- [ ] **Convert FHIR values to CQL values**: map each input `parameter` to a CQL value keyed by `parameter.name`, following the *Using CQL* type mapping. Honor the repetition/parts semantics from the spec:
  - a parameter that appears more than once → a CQL `List`,
  - a parameter with `parameter.part` → a CQL `Tuple`.
- [ ] **Merge over defaults** in `enhance-context`: build the effective parameter map as `(merge parameter-default-values supplied-parameters)` and pass that to both `:parameters` and `library/resolve-params`, so supplied values override declared defaults while undeclared/missing values fall back to the library default.
- [ ] **Specs**: extend the params spec (`evaluate_measure/spec.clj` / `measure/spec.clj`) so the new `:parameters` entry is covered, with the corresponding `s/fdef`s in the `*-spec` namespace.
- [ ] **Documentation**: document the new input in `docs/api/operation/measure-evaluate-measure.md` and the CQL query guide `docs/cql-queries/api.md`, noting it is POST-only.

### Notes / open questions

- **POST only**: the input is a `Parameters` resource, so it is supported solely via `POST` with a `Parameters` body. It has no `GET` representation.
- **Interaction with `periodStart`/`periodEnd`**: Blaze derives the measurement period from `periodStart`/`periodEnd` (`:period`), independent of CQL parameters. Decide how an explicitly supplied `Measurement Period` parameter should interact with these (ignore, error, or override) and document the chosen behavior.
- **Unknown parameters**: decide whether supplying a value for a name the library does not declare is silently ignored or returns an anomaly.
- **Type coverage / phasing**: the full FHIR→CQL type mapping is broad. A first phase could cover the common primitive types (date, dateTime, integer, decimal, boolean, string, code) plus the List/Tuple rules, with complex types (Quantity, Period, Coding/CodeableConcept, Ratio, …) following up. Any unsupported type should return a clear Cognitect anomaly rather than silently dropping the value.

### Acceptance Criteria

- [ ] A `POST …/Measure/$evaluate-measure` with a `parameters` input overrides the matching CQL library parameter, changing the evaluation result accordingly.
- [ ] A parameter repeated in the input becomes a CQL `List`; a parameter with parts becomes a CQL `Tuple`.
- [ ] Library defaults still apply for parameters not supplied in the request.
- [ ] Unsupported parameter types / unknown names produce an appropriate anomaly (no silent failures).
- [ ] Public functions have specs (`s/fdef` in the `*-spec` namespace; data specs in the `*.spec` namespace).
- [ ] Tests written TDD-first, with spec instrumentation and the standard `tu/fixture`; coverage ≥ 95% forms.
- [ ] Documentation updated.

## References

- R5 `$evaluate-measure` operation (`parameters` input): https://hl7.org/fhir/R5/measure-operation-evaluate-measure.html
- R4 `$evaluate-measure` operation (no `parameters`): https://hl7.org/fhir/R4/operation-measure-evaluate-measure.html
- Param parsing: `modules/operation-measure-evaluate-measure/src/blaze/fhir/operation/evaluate_measure/middleware/params.clj`
- Context build / parameter resolution: `modules/operation-measure-evaluate-measure/src/blaze/fhir/operation/evaluate_measure/measure.clj`
- Library parameter defaults & `resolve-params`: `modules/cql/src/blaze/elm/compiler/library.clj`

[1]: <https://hl7.org/fhir/R5/measure-operation-evaluate-measure.html>
[2]: <https://hl7.org/fhir/R4/operation-measure-evaluate-measure.html>

alexanderkiel:
Closes: #3745

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01QKf7QCm2d2zDfGyTLUKq5t

### DIFF
--- a/docs/api/operation/measure-evaluate-measure.md
+++ b/docs/api/operation/measure-evaluate-measure.md
@@ -10,6 +10,25 @@ The official documentation can be found [here][1].
 * a detailed documentation how to use the $evaluate-measure API can be found [here](../../cql-queries/api.md)
 * a documentation how to use $evaluate-measure via blazectl can be found [here](../../cql-queries/blazectl.md) 
 
+## Parameters Input
+
+In addition to the parameters defined by the R4 operation, Blaze supports the [`parameters`][4] input that was added to the operation in FHIR R5. It allows a client to pass values into the CQL evaluation, overriding the default values declared in the measure's library by name. This way a single measure can be evaluated with different inputs (e.g. thresholds, codes, date boundaries) without editing and re-publishing the library.
+
+Because the `parameters` input is a `Parameters` resource nested in `parameter.resource`, it can **only** be supplied via `POST` with a `Parameters` body. It has no representation in the `GET` query string.
+
+The following rules apply:
+
+* a supplied parameter overrides the matching CQL library parameter; parameters not supplied fall back to the library default
+* a parameter that appears more than once becomes a CQL `List`
+* a parameter with `parameter.part` becomes a CQL `Tuple`
+* supplying a value for a name the library doesn't declare returns an error
+* the measurement period is derived from `periodStart`/`periodEnd` independently of CQL parameters; a supplied `Measurement Period` parameter only affects the CQL evaluation and does not change the period of the resulting `MeasureReport`
+
+Currently the common primitive types (`boolean`, `integer`, `decimal`, `string`, `code`, `date` and `dateTime`) together with the `List`/`Tuple` rules are supported. Unsupported types return an error.
+
+A detailed example can be found in the [CQL Evaluation API documentation](../../cql-queries/api.md#evaluation-with-parameters).
+
 [1]: <https://hl7.org/fhir/R4/operation-measure-evaluate-measure.html>
 [2]: <https://github.com/samply/blazectl>
 [3]: <https://github.com/samply/blazectl/releases/tag/v0.15.0>
+[4]: <https://hl7.org/fhir/R5/measure-operation-evaluate-measure.html>
--- a/docs/cql-queries/api.md
+++ b/docs/cql-queries/api.md
@@ -154,6 +154,39 @@ the result should be the following [MeasureReport][6]:
 
 Under `group[0].population[0].count` the result of the query can be found.
 
+## Evaluation with Parameters
+
+A CQL library can declare parameters that are used in its expressions, optionally with a default value:
+
+```text
+library Retrieve
+using FHIR version '4.0.0'
+include FHIRHelpers version '4.0.0'
+
+parameter "Gender" String default 'male'
+
+context Patient
+
+define InInitialPopulation:
+  Patient.gender = "Gender"
+```
+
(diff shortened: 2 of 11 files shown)


alexanderkiel:
Handled by #3748

PR #3836, opened 2026-07-09
Toggle Summary Mode in UI

### LINKED ISSUE #2812: Toggle Summary Mode in the UI
We may even offer a toggle to enable/disable the summary mode that will be enabled by default. On top of that the UI could remember that toggle per resource type in local storage. But we have to check that agains server side rendering first. We may also store user settings inside the node backend. So we don't have to do that in Blaze itself.

knoppiks:
Closes: #2812

### DIFF
--- a/modules/frontend/src/lib/resource/subsetted.ts
+++ b/modules/frontend/src/lib/resource/subsetted.ts
@@ -1,0 +1,28 @@
+import type { Bundle, BundleEntry } from 'fhir/r4';
+import type { SummaryMode } from '$lib/summary.js';
+
+const subsettedSystem = 'http://terminology.hl7.org/CodeSystem/v3-ObservationValue';
+
+/**
+ * Whether the resource of the given bundle entry is a subsetted (summary)
+ * representation, according to its SUBSETTED meta tag.
+ */
+export function isSubsetted(entry: BundleEntry): boolean {
+  const tags = entry.resource?.meta?.tag;
+  return tags?.filter((c) => c.system === subsettedSystem)[0]?.code === 'SUBSETTED';
+}
+
+/**
+ * The summary mode a bundle's resources reveal, or undefined if the bundle
+ * gives no evidence of its summary mode.
+ *
+ * Bundles without any resource, like `_summary=count` results or history
+ * bundles of deletions only, give no evidence of their summary mode. Count is
+ * never inferred, because an empty bundle is indistinguishable from an empty
+ * result set.
+ */
+export function bundleSummaryMode(bundle: Bundle): SummaryMode | undefined {
+  const entries = bundle.entry?.filter((entry) => entry.resource !== undefined) ?? [];
+  if (entries.length === 0) return undefined;
+  return entries.some(isSubsetted) ? 'summary' : 'full';
+}
--- a/modules/frontend/src/lib/util.ts
+++ b/modules/frontend/src/lib/util.ts
@@ -70,7 +70,7 @@ export function moveDownAtIndex<T>(array: T[], i: number): T[] {
 
 export const defaultCount = '20';
 
-export function processParams(params: URLSearchParams): string {
+export function processParams(params: URLSearchParams, summary = true): string {
   const newParams = new URLSearchParams();
   for (const [name, value] of params) {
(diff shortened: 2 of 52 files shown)


alexanderkiel (changes requested):
[Summary Mode UX Review.pdf](https://github.com/user-attachments/files/29893713/Summary.Mode.UX.Review.pdf)

### UI Proposal

Screenshot 2026-07-10 at 16 53 09

knoppiks:
#3994 found while implementing. Rebase after #3996 lands

modules/frontend/src/routes/__history-page/[pageId=pageId]/+page.svelte
```diff
@@ -19,8 +25,11 @@
         <TotalBadge total={data.bundle.total} />
       {/if}
     </p>
+    <SummaryControl enabled={summaryEnabled} fixed />
```
Comment (alexanderkiel) [inline modules/frontend/src/routes/__history-page/[pageId=pageId]/+page.svelte]: Just omitting type here sets the default. I think, we need a history setting that is independend from default. Also think about separating the kind `history` from kind `search`.

modules/frontend/src/lib/resource/subsetted.ts
```diff
@@ -0,0 +1,22 @@
...
+  return tags?.filter((c) => c.system === subsettedSystem)[0]?.code === 'SUBSETTED';
+}
+
+/**
+ * Whether the given bundle contains subsetted resources.
+ *
+ * Used to tell the summary mode of a result set that was returned by an opaque
+ * page link and so can't be derived from the request.
+ */
+export function isBundleSubsetted(bundle: Bundle): boolean {
```
alexanderkiel: Should return undefined if there is no evidence of subsetted. For example if _summary=count and we have no entries. Or on delete only history bundles.

alexanderkiel:
Instance history should have also summary control. But summary should be always default to true and being not remembered in the cookie. So users can disable summary temporarily but a page reload will enable it again. That's not nice, but we can't store summary settings for each instance in the cookie.

alexanderkiel:
PerType cookie entries that equal default should be pruned to save cookie size.

modules/frontend/src/lib/util.ts
```diff
@@ -81,8 +81,9 @@ export function processParams(params: URLSearchParams): string {
   if (!newParams.has('_count')) {
     newParams.append('_count', defaultCount);
   }
-  // we like to have summary representations by default
-  if (!newParams.has('_summary')) {
+  // request summary representations when summary mode is enabled; omitting the
+  // param yields full resources
+  if (summaryEnabled && !newParams.has('_summary')) {
```
alexanderkiel: In case we do not enable summary mode, because we have already _summary=count, we should communicate that upstream so that summaryEnabled is no longer true.

alexanderkiel:
If we are in _summary=count, no summary control and no banner should be displayed.

alexanderkiel (changes requested):

modules/frontend/src/lib/resource/subsetted.ts
```diff
@@ -0,0 +1,26 @@
...
+ * The summary state of a result set that was returned by an opaque page link
+ * and so can't be derived from the request.
+ *
+ * Bundles without any resource, like `_summary=count` results or history
+ * bundles of deletions only, give no evidence of their summary mode. The mode
+ * of a page is fixed, because the page link determines it.
+ */
+export function bundleSummaryState(bundle: Bundle): SummaryState {
+  const entries = bundle.entry?.filter((entry) => entry.resource !== undefined) ?? [];
+  return entries.length === 0 ? {} : { enabled: entries.some(isSubsetted), fixed: true };
```
alexanderkiel: Should it `{fixed: true}` in case we have no entries?

alexanderkiel (changes requested):

modules/frontend/src/lib/resource/subsetted.ts
knoppiks: the fn was only used on subsequent history pages - therefor `fixed: true` was always needed. I refactored it to make that clear.

alexanderkiel (approved):

PR #3838, opened 2026-07-09
Support Profile Validation via External Validator

### LINKED ISSUE #3697: Support Profile Validation via External Validator
Implement support for an external FHIR resource validation mechanism in Blaze configured via environment variables.
Must work with https://github.com/medizininformatik-initiative/mii-fhir-validator

Introduce `EXTERN_VALIDATOR_URL`. When this environment variable is set to a valid URL, forward all incoming resources to the designated external validator endpoint for schema and profile validation prior to persistence.

Introduce a configuration switch to control the behavior when a resource fails external validation.
* Tagging Mode (Default): Accept and store the invalid resource, but flag or tag it as invalid.
* Rejection Mode: Block the invalid resource entirely and return an appropriate HTTP error response (e.g., 400 Bad Request with an accompanying OperationOutcome).

alexanderkiel:
Closes: #3697


modules/extern-validator/src/blaze/validator/extern/semaphore.clj
knoppiks, line 1: Should we guard against permit inflation here? A release! with no pending waiters increments the permit count unconditionally, so a spurious or double release! (one without a matching acquire!) raises permits above the initial value and permanently weakens the concurrency bound.

modules/extern-validator/src/blaze/validator/extern/impl.clj
```diff
@@ -0,0 +1,77 @@
...
+    (seq (:security meta)) (assoc :security (:security meta))
+    (seq (:tag meta)) (assoc :tag (:tag meta))
+    (seq (:extension meta)) (assoc :extension (:extension meta))))
+
+(defn tag-invalid
+  "Returns `resource` tagged as invalid.
+
+  When `contain-outcome?` is true, additionally embeds `operation-outcome` as a
+  contained resource referenced from a meta extension."
+  [resource operation-outcome contain-outcome?]
```
knoppiks: `contain-outcome` sounds like a predicate, could easily misread with `contains-outcome`.
```suggestion
  [resource operation-outcome embed-outcome?]
```

modules/extern-validator/test/blaze/validator/extern/impl_test.clj
```diff
@@ -0,0 +1,109 @@
...
+    (fn [severity]
+      {:fhir/type :fhir.OperationOutcome/issue
+       :severity (type/code severity)
+       :code #fhir/code "processing"
+       :diagnostics (type/string (str severity " issue"))})
+    severities)})
+
+(deftest invalid?-test
+  (testing "an error issue is invalid"
+    (is (true? (impl/invalid? (operation-outcome "error")))))
```
knoppiks: is `true?` necessary? The spec already ensures a bool return value.

knoppiks (changes requested):

modules/extern-validator/src/blaze/validator/extern/semaphore.clj
alexanderkiel: The JDK `java.util.concurrent.Semaphore` has the same issue. But it's by design. In the documentation, the sentence: "Correct usage of a semaphore is established by programming convention in the application." is used.

knoppiks (dismissed):

knoppiks (dismissed):

knoppiks (approved):

PR #3894, opened 2026-07-17
Batch Local Tx Log Writes and Add Backpressure

### LINKED ISSUE #3886: Batch Local Tx Log Writes and Add Backpressure
Currently the local transaction log serializes `-submit` calls with a lock, blocking the calling thread while another submit stores its transaction data. Each transaction is stored in the key-value store individually and submits are accepted without any limit.

Rework the local transaction log to keep its complete state in a single atom:

* `:t` — the last assigned point in time; `-submit` assigns `t` and buffers the transaction data in one `swap!`, without blocking the calling thread
* `:stored-t` — the highest `t` up to which transaction data is durably stored, gaps of failed writes included; pollers only ever see transaction data up to `:stored-t`
* `:buffer` — a sorted map of `t` to buffered transaction data, including any local payload: entries not yet stored plus stored entries retained until the poller acknowledges them via the `poll!` offset (#3887)
* `:signal` — a CompletableFuture that is completed and renewed whenever `:stored-t` advances, waking up blocked pollers
* `:store-signal` — a CompletableFuture that is completed on each submit and renewed by the store loop, waking it up to store new transaction data
* `:closed?` — true once the transaction log is closed, terminating the store loop and rejecting further submits with an unavailable anomaly

A single storer thread (named `<main|admin>-local-tx-log-storer`) takes all unstored entries from the buffer and stores them with one batched `kv/put!` (group commit). If storing fails, the affected submit futures complete exceptionally and `:stored-t` advances past the failed entries, leaving a gap in `t` (gaps in `t` are allowed, as before). Storing never throws, so an unexpected error can't terminate the store loop and leave submitters waiting on futures that are never completed. `-last-t` continues to report the last assigned `t`. A new `store-batch-entries` histogram shows how many submits are coalesced into one batch. Pollers that are behind the buffer read from storage as before.

The buffer itself is not bounded. Backpressure is applied one layer up, in the database node, because the transaction log accepts whatever is submitted to it and the node is the only place a transaction can be turned away before anything is written. The node tracks the transactions that are submitted but not yet indexed in its own state:

* `:submitting` — the number of transactions accepted for submission whose `t` isn't known yet
* `:in-flight` — a sorted set of the `t`s already assigned by the log but not yet indexed

A submit that finds `:submitting` plus `:in-flight` already at `DB_MAX_IN_FLIGHT_TRANSACTIONS` (default 1024) is rejected with a busy anomaly (HTTP 503) before any resource content is stored, and is counted in a new `submit-rejections-total` counter. A place is freed once the transaction was indexed, or right away if the submit never reached the transaction log. Because the poller acknowledges everything it has indexed, that limit is what bounds the buffer. Enforcing it per node instead of per buffer also bounds how far a node can run ahead of its own indexing with a Kafka transaction log.

Depends on #3887.

alexanderkiel:
Closes: #3886


alexanderkiel:
Load test results, baseline (main, `03855d9b1`) vs. this branch, both with `DB_RESOURCE_STORE_KV_THREADS=64`:

| System | VUs | Req/s before | Req/s after | Speedup | q99 before | q99 after |
|--------|----:|-------------:|------------:|--------:|-----------:|----------:|
| A5N46  |  32 |        145.7 |      1000.0 |    6.9× |     693 ms |     42 ms |
| A5N46  |  64 |        162.6 |      2064.3 |   12.7× |    1781 ms |     42 ms |
| LEA47  |  32 |        443.6 |      1392.5 |    3.1× |     218 ms |     31 ms |
| LEA47  |  64 |        452.2 |      1370.3 |    3.0× |     434 ms |     68 ms |
| LEA79  |  64 |       5617.3 |      5381.2 |   0.96× |    15.6 ms |   13.0 ms |

The gain tracks fsync latency: on A5N46 (5.4 ms syncs) the baseline serialized one tx-log fsync per transaction, while batching amortizes it across everything buffered in flight — the median holds at ~30 ms from 4 to 64 clients and throughput still doubles with each doubling of clients at 64, i.e. unsaturated. LEA79 syncs in 8 µs, was never fsync-bound, and stays at parity (−4% at 64 VUs, within noise).

knoppiks (approved):

Issue #3897, opened 2026-07-19
Quarantine Flaky medication-administration-mem-size-test


alexanderkiel:
## Summary

`blaze.fhir.spec-test/medication-administration-mem-size-test` (`modules/fhir-structure/test/blaze/fhir/spec_test.clj:6971`) is flaky in the `test-mem-size (large-heap)` CI job. It is a `test.check` property test asserting that `Base/memSize` (a hand-rolled, allocation-free estimate of a FHIR resource's retained heap size) is within 10% of the real measured size obtained via JOL (`blaze.fhir.spec.memory/total-size`, `org.openjdk.jol`).

Observed in the CI flakiness scan for 2026-07-12 to 2026-07-19:
- FAIL: https://github.com/samply/blaze/actions/runs/29486704401/job/87582915906 (push, commit `5e94170548561d3686ae8357ad657e2dbf5b354b`, 2026-07-16T09:19–10:24 UTC — job ran ~65 minutes before failing)
- PASS on the **same commit**: https://github.com/samply/blaze/actions/runs/29485435600/job/87579101096 (merge_group, ~20 minutes earlier)
- Failure: `64 tests, 119 assertions, 2 failures.` — property fails on a `test.check`-generated `MedicationAdministration`, `:seed 1784197479734`, shrinks to a minimal resource dominated by very short string fields (extension URLs, codes, systems ≤ 4 chars).

## Root-cause hypothesis (not fully confirmed — see below)

`blaze.fhir.spec.type.String.Interned#memSize()` unconditionally returns `0` (`String.java:260-262`) for short (≤4 char) interned strings, on the assumption that their cost is amortized/shared elsewhere (`blaze.Interners.weakInterner`, `String.java:189`). `Base/memSize` (the "calculated" side) applies this 0-cost assumption per instance regardless of resource size.

The property's "real size" side measures the *marginal* cost of a second parse of the same bytes via a JOL graph-layout diff (`total-size`), which only reflects `0` for those same fields if the interner really does hand back the identical instance both times.

[...]

Issue #3891, opened 2026-07-17
Fix Flaky Page Store Token Generation Test


alexanderkiel:
## Summary

`blaze.page-store.token-test/generate-test` failed on CI run [29592845826](https://github.com/samply/blaze/actions/runs/29592845826/job/87926373971) with:

```
Exception: clojure.lang.ExceptionInfo: Call to blaze.page-store.local.hash/hash-clause did not conform to spec.
{... :clojure.spec.alpha/value ([]) ... :clojure.spec.alpha/failure :instrument ...}
```

The property generated the clauses value `[[]]` — a disjunction that is an *empty* vector of clauses — and the instrumented `hash-clause` rejected `[]` because it is neither a valid search clause nor a valid sort clause.

## Root cause

Two defects combine to produce the flakiness:

1. **The spec is too loose.** The `:clauses` branch of `:blaze.db.query/disjunction` is `(s/coll-of :blaze.db.query/clause :kind vector?)` without `:min-count 1`, so the derived generator produces empty disjunctions (about 20 % of sampled `:blaze.db.query/clauses` values contain one). An empty disjunction is semantically meaningless and unsupported: `blaze.page-store.token/hash-disjunction` misclassifies `[]` as a single clause and passes it to `hash-clause`, whose args spec rejects it. The internal spec in `blaze.db.impl.index-spec` already requires `:min-count 1` for disjunctions.

2. **The test namespace doesn't follow the standard fixture convention.** `blaze.page-store.token-test` neither calls `(st/instrument)` at the top level nor uses `tu/fixture`, so whether `hash-clause` is instrumented during `generate-test` depends on kaocha's randomized test order: sibling namespaces instrument at load time, and the first `tu/fixture`-using test to run unstruments everything afterwards. Only when `generate-test` is scheduled before any fixture-using test does it run instrumented and hit the empty-disjunction crash — hence the sporadic failures.

## Fix

* Add `:min-count 1` to the `:clauses` branch of `:blaze.db.query/disjunction` in `modules/spec/src/blaze/spec.clj`, so an empty disjunction is invalid and is never generated.
* Bring `blaze.page-store.token-test` in line with the standard test conventions (top-level `(st/instrument)` and `tu/fixture`), which makes the previously order-dependent failure deterministic before the spec fix and keeps the test permanently instrumented after it.

## Verification

[...]

PR #3898, opened 2026-07-19
Quarantine Flaky medication-administration-mem-size-test

alexanderkiel:
## Summary

Weekly CI flakiness scan (2026-07-12 to 2026-07-19) found `blaze.fhir.spec-test/medication-administration-mem-size-test` failing on a commit where it also passed (same SHA, different runs). Two other flaky tests found in the same scan were already fixed by earlier commits:

- `blaze.db.node.tx-indexer.verify-test/verify-tx-cmds-test` — fixed by 5b95f5a
- `blaze.page-store.token-test` — fixed by 030e232 (#3891)

This is the one remaining unfixed flake.

## Root cause

`blaze.fhir.spec.type.String.Interned#memSize()` returns `0` for short interned strings under the assumption their cost is amortized elsewhere. `Base/memSize`'s "calculated size" applies that assumption per-instance. The property test compares this against a JOL-measured "real size" with a 10%-relative tolerance. For a `test.check`-shrunk *minimal* counterexample dominated by several short strings, the fixed per-string omission can exceed 10% of the (very small) total, even though it's negligible for realistic resources.

I could not get a reliable local repro to confirm this and safely fix the underlying calculation under this repo's mandatory TDD workflow (write a failing test, then fix): JOL's `Instrumentation` agent can't attach in the sandbox this was investigated in (`Unable to get Instrumentation. Dynamic Attach failed`), so the "real size" measurement there doesn't match what the actual `test-mem-size (large-heap)` CI runner sees — re-running with the exact recorded seed against the exact failing commit passed locally.

Full analysis: #3897

## Changes

Quarantines the test with `^:kaocha/pending` (skip, reported as pending — not run) and a comment linking #3897, per the flaky-test playbook: fix if confident, quarantine if not.

## Verification

- `make -C modules/fhir-structure fmt` — passes
- `make -C modules/fhir-structure lint` — passes
- `make -C modules/fhir-structure test` — 220 tests, 4561 assertions, 0 failures
- `make -C modules/fhir-structure test-mem-size-large-heap` (the job that was flaky) — 64 tests, 118 assertions, **1 pending**, 0 failures

## Test plan

- [x] Confirm the target test is skipped and reported `PENDING`, not silently vanished
- [x] Confirm the rest of the `test-mem-size-large-heap` suite (63 other mem-size tests) still passes
- [x] Confirm the regular module test suite is unaffected
- [ ] Follow-up (tracked in #3897): reproduce with real JOL instrumentation (on CI or a machine that allows agent attach) and land a real fix — widen tolerance with an absolute floor, or stop assuming 0 cost for interned strings, or bound generator size.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CDCkYRndNpFzjWQrvGDvcj)_

### DIFF
--- a/modules/fhir-structure/test/blaze/fhir/spec_test.clj
+++ b/modules/fhir-structure/test/blaze/fhir/spec_test.clj
@@ -6967,7 +6967,10 @@
       (prop/for-all [medication-administration (fg/medication-administration)]
         (murmur3 medication-administration)))))
 
-(deftest ^:mem-size medication-administration-mem-size-test
+;; Flaky: the calculated size occasionally falls more than 10% short of the
+;; real size on minimal shrunk counterexamples. See
+;; https://github.com/samply/blaze/issues/3897 for the root-cause analysis.
+(deftest ^:mem-size ^:kaocha/pending medication-administration-mem-size-test
   (satisfies-prop 20
     (memsize-prop "medication-administration")))
 


alexanderkiel:
I can't find a solution for this problem.

PR #3911, opened 2026-07-21
Add Date and DateTime Tests to Interval Boundary Operators

### LINKED ISSUE #3908: Add Date and DateTime Tests to Interval Boundary Operators
Several interval operators in `blaze.elm.compiler.interval-operators-test` only cover the `Integer` and `Decimal` point types. Add `Date` and `DateTime` test coverage as well.

All of these operators rely on the semantics described in the `Start` and `End` operators to determine interval boundaries — in particular, null bounds resolve to the minimum/maximum value of the point type at interval construction.

Point operators (return the resolved boundary):

- `compile-end-test` — closed high → high value; open high (`:>`) → predecessor; null high → maximum value
- `compile-start-test` — closed low → low value; open low (`:<`) → successor; null low → minimum value

Binary boundary predicates — add `Date`/`DateTime` blocks exercising closed intervals with null bounds (resolving to the minimum/maximum value), covering both matching and non-matching cases:

- `compile-overlaps-test`
- `compile-ends-test`
- `compile-starts-test`
- `compile-meets-before-test`
- `compile-meets-after-test`
- `compile-includes-test`

`IncludedIn`, `OverlapsBefore` and `OverlapsAfter` are normalized to other operators and have no direct interval test bodies, so they are covered transitively.

This is a test-only change; the production behaviour of the operators is unaffected.

alexanderkiel:
Closes: #3908

### DIFF
--- a/modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
+++ b/modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
@@ -9,6 +9,7 @@
    [blaze.elm.compiler.core-spec]
    [blaze.elm.compiler.interval-operators]
    [blaze.elm.compiler.test-util :as ctu :refer [has-form]]
+   [blaze.elm.date-time :as date-time]
    [blaze.elm.decimal :as decimal]
    [blaze.elm.interval :refer [interval]]
    [blaze.elm.literal :as elm]
@@ -606,6 +607,20 @@
       [#elm/decimal "1" #elm/decimal "2.1" :>] 2.09999999M
       [#elm/decimal "1" {:type "Null"}] decimal/max))
 
+  (testing "Date"
+    (are [x res] (= res (ctu/compile-unop elm/end elm/interval x))
+      [#elm/date "2019" #elm/date "2021"] #system/date"2021"
+      [#elm/date "2019" #elm/date "2021" :>] #system/date"2020"
+      [#elm/date "2019" {:type "Null"}] date-time/max-date
+      [#elm/date "2019" {:type "Null"} :>] nil))
+
+  (testing "DateTime"
+    (are [x res] (= res (ctu/compile-unop elm/end elm/interval x))
+      [#elm/date-time "2019" #elm/date-time "2021"] #system/date-time"2021"
+      [#elm/date-time "2019" #elm/date-time "2021" :>] #system/date-time"2020"
+      [#elm/date-time "2019" {:type "Null"}] date-time/max-date-time
+      [#elm/date-time "2019" {:type "Null"} :>] nil))
+
   (ctu/testing-unary-null elm/end)
 
   (ctu/testing-unary-op elm/end)
@@ -643,6 +658,20 @@
         [#elm/integer "1" #elm/parameter-ref "3"] [#elm/integer "1" #elm/parameter-ref "3"] true?
         [#elm/integer "2" #elm/parameter-ref "3"] [#elm/integer "1" #elm/parameter-ref "3"] true?
         [#elm/integer "1" #elm/parameter-ref "3"] [#elm/integer "2" #elm/parameter-ref "3"] false?)))
+
+  (testing "Date"
+    (are [x y pred] (pred (ctu/compile-binop elm/ends elm/interval x y))
+      [#elm/date "2019" {:type "Null"}] [#elm/date "2018" {:type "Null"}] true?
+      [#elm/date "2019" {:type "Null"} :>] [#elm/date "2018" {:type "Null"} :>] nil?
+      [{:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}Date"} #elm/date "2021"] [{:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}Date"} #elm/date "2021"] true?
+      [#elm/date "2019" {:type "Null"}] [#elm/date "2020" {:type "Null"}] false?))
+
+  (testing "DateTime"
+    (are [x y pred] (pred (ctu/compile-binop elm/ends elm/interval x y))
+      [#elm/date-time "2019" {:type "Null"}] [#elm/date-time "2018" {:type "Null"}] true?
+      [#elm/date-time "2019" {:type "Null"} :>] [#elm/date-time "2018" {:type "Null"} :>] nil?
+      [{:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}DateTime"} #elm/date-time "2021"] [{:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}DateTime"} #elm/date-time "2021"] true?
+      [#elm/date-time "2019" {:type "Null"}] [#elm/date-time "2020" {:type "Null"}] false?))
 
   (ctu/testing-binary-null elm/ends interval-zero)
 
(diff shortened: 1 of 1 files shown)


modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
```diff
@@ -606,6 +607,18 @@
       [#elm/decimal "1" #elm/decimal "2.1" :>] 2.09999999M
       [#elm/decimal "1" {:type "Null"}] decimal/max))
 
+  (testing "Date"
+    (are [x res] (= res (ctu/compile-unop elm/end elm/interval x))
+      [#elm/date "2019" #elm/date "2021"] #system/date"2021"
+      [#elm/date "2019" #elm/date "2021" :>] #system/date"2020"
```
paulolaup, line 613: If `end` returns `predecessor of <highBoundary>` why would that be `2020`?
`Date(2020, 07, 01) < Date(2021)` should be `true` according to the spec. So shouldn't the predecessor of `2021` be `2020-12-31` (when handling dates)?
Are intervals members restricted to the granularity of their boundaries?

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
```diff
@@ -606,6 +607,18 @@
...
+      [#elm/date "2019" {:type "Null"}] date-time/max-date))
+
+  (testing "DateTime"
+    (are [x res] (= res (ctu/compile-unop elm/end elm/interval x))
+      [#elm/date-time "2019" #elm/date-time "2021"] #system/date-time"2021"
+      [#elm/date-time "2019" #elm/date-time "2021" :>] #system/date-time"2020"
+      [#elm/date-time "2019" {:type "Null"}] date-time/max-date-time))
```
paulolaup: Add test case for open interval with `null`-valued boundary

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
```diff
@@ -783,6 +808,18 @@
         [#elm/integer "1" #elm/integer "2"] [#elm/integer "1" #elm/integer "2"] true?
         [#elm/integer "1" #elm/integer "2"] [#elm/integer "1" #elm/integer "3"] false?))
 
+    (testing "Date"
+      (are [x y pred] (pred (ctu/compile-binop elm/includes elm/interval x y))
+        [#elm/date "2019" {:type "Null"}] [#elm/date "2020" {:type "Null"}] true?
```
paulolaup, line 817: What about `Interval[@2019, null] includes Interval[@2020, null)`? Does the first include the second since the high boundary is the max value and the high boundary of the second is an unknown value between `2020` and the max value (unless negative intervals are allowed)?

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
```diff
@@ -877,6 +914,16 @@
       [#elm/integer "1" #elm/integer "2"] [#elm/integer "3" #elm/integer "4"] true?
       [#elm/integer "1" #elm/integer "2"] [#elm/integer "4" #elm/integer "5"] false?))
 
+  (testing "Date"
+    (are [x y pred] (pred (ctu/compile-binop elm/meets-before elm/interval x y))
+      [{:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}Date"} #elm/date "2018"] [#elm/date "2019" {:type "Null"}] true?
+      [{:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}Date"} #elm/date "2018"] [#elm/date "2020" {:type "Null"}] false?))
+
```
paulolaup, line 929: Add tests to ensure `Interval[@2019, null] meets before Interval[null, @2020]` is always `false`.

How about `Interval[@2019, @2020] meets before Interval(null, @2021]`? Would this return `false` or `null`?

paulolaup (changes requested):
Improve completeness of tests, e.g. cover all scenarios with each test (if applicable) so that we know what happens for intervals with unknown inclusive/exclusive boundaries

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
alexanderkiel: From the CQL Logical Specification:

> For Date, DateTime, and Time values, Predecessor is equivalent to subtracting a time-unit quantity for the lowest specified precision of the value. For example, if the DateTime is fully specified, Predecessor is equivalent to subtracting 1 millisecond; if the DateTime is specified to the second, Predecessor is equivalent to subtracting one second, etc.

https://cql.hl7.org/04-logicalspecification.html

So `2020` is correct.

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
alexanderkiel: added the test cases to start and end operators in date and dateTime

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
alexanderkiel: `Interval[@2019, null] includes Interval[@2020, null)` returns `nil` because the upper bound of `Interval[@2020, null)` is unknown. Although the unknown upper bound can't be greater than the max value, it doesn't work this way. The `End` operator will return `nil` for `Interval[@2020, null)` and `GreaterOrEqual` will return `nil` whenever one argument is `nil`. So the answer is `nil`. I added such tests.

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
alexanderkiel: `Interval[@2019, null] meets before Interval[null, @2020]` will return `nil` because of `Predecessor` of the minimum value returns `nil`:

> If the argument is already the minimum value for the type, a null is returned.

https://cql.hl7.org/04-logicalspecification.html#predecessor

`Interval[@2019, @2020] meets before Interval(null, @2021]` will also return `nil` because the lower bound of `Interval(null, @2021]` is `nil`.

Added test cases.

paulolaup: I don't entirely get why the first case results in `nil` instead of `false` since IMO the semantics of the operator would imply such. Or at least, logically speaking, I would consider the question posed by this arrangement to be answerable.

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
```diff
@@ -644,6 +659,18 @@
         [#elm/integer "2" #elm/parameter-ref "3"] [#elm/integer "1" #elm/parameter-ref "3"] true?
         [#elm/integer "1" #elm/parameter-ref "3"] [#elm/integer "2" #elm/parameter-ref "3"] false?)))
 
+  (testing "Date"
```
paulolaup, line 662: (Applies to both tests `Date` and `DateTime`): What about cases where the interval has an exclusive `null`-valued (high) boundary?

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
```diff
@@ -877,6 +918,18 @@
...
+      [{:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}Date"} #elm/date "2018"] [#elm/date "2019" {:type "Null"}] true?
+      [{:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}Date"} #elm/date "2018"] [#elm/date "2020" {:type "Null"}] false?
+      [#elm/date "2019" {:type "Null"}] [{:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}Date"} #elm/date "2020"] nil?
+      [#elm/date "2019" {:type "Null"}] [:< {:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}Date"} #elm/date "2021"] nil?))
+
+  (testing "DateTime"
+    (are [x y pred] (pred (ctu/compile-binop elm/meets-before elm/interval x y))
+      [{:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}DateTime"} #elm/date-time "2018"] [#elm/date-time "2019" {:type "Null"}] true?
+      [{:type "Null" :resultTypeName "{urn:hl7-org:elm-types:r1}DateTime"} #elm/date-time "2018"] [#elm/date-time "2020" {:type "Null"}] false?))
+
```
paulolaup, line 936: Add the last two test cases of test `Date` to test `DateTime` as well

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
```diff
@@ -901,6 +954,16 @@
       [#elm/integer "3" #elm/integer "4"] [#elm/integer "1" #elm/integer "2"] true?
       [#elm/integer "4" #elm/integer "5"] [#elm/integer "1" #elm/integer "2"] false?))
 
+  (testing "Date"
```
paulolaup, line 961: Missing tests with relevant where relevant boundaries have `null` values.

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
```diff
@@ -1172,6 +1263,18 @@
       [#elm/integer "1" #elm/integer "2"] [#elm/integer "1" #elm/integer "3"] true?
       [#elm/integer "2" #elm/integer "3"] [#elm/integer "1" #elm/integer "3"] false?))
 
+  (testing "Date"
```
paulolaup, line 1274: Missing tests with relevant where relevant boundaries have `null` values. The tests for operator `start` have them

paulolaup (changes requested):

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
alexanderkiel: done

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
alexanderkiel: applied inverse `ends` operator tests

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
alexanderkiel: done

modules/cql/test/blaze/elm/compiler/interval_operators_test.clj
alexanderkiel: done

paulolaup (approved):

PR #3946, opened 2026-07-27
Remove Sleeps From Executor Shutdown Timeout Tests

### LINKED ISSUE #3871: Remove Sleeps From Executor Shutdown Timeout Tests
Part of the effort to eliminate all `Thread/sleep` calls from tests (see #3863 and PR #3866).

Six tests share the same pattern: submit `#(Thread/sleep 11000)` so that `ig/halt!` runs into the 10 s termination timeout, plus a `(Thread/sleep 100)` to "ensure that the function is called" before halting. Both sleeps can be replaced by a task that blocks on a never-released `CountDownLatch` and a second latch signalling that the task has started. This is deterministic and shaves ~100 ms of real time per test:

* `modules/cql/test/blaze/elm/expression/cache_test.clj:112,115` (`executor-shutdown-timeout-test`)
* `modules/db/test/blaze/db/node_test.clj:331,334` (`indexer-executor-shutdown-timeout-test`)
* `modules/db-resource-store/test/blaze/db/resource_store/kv_test.clj:292,295` (`executor-shutdown-timeout-test`)
* `modules/db-tx-log-kafka/test/blaze/db/tx_log/kafka_test.clj:243,246` (`last-t-executor-shutdown-timeout-test`)
* `modules/operation-measure-evaluate-measure/test/blaze/fhir/operation/evaluate_measure_test.clj:1073,1076` (`indexer-executor-shutdown-timeout-test`)
* `modules/scheduler/test/blaze/scheduler_test.clj:51,55` (`shutdown-timeout-test`)

Consider a shared helper in a test-util module since the pattern is identical everywhere.

trobanga:
Closes #3871

### DIFF
--- a/modules/test-util/src/blaze/test_util.clj
+++ b/modules/test-util/src/blaze/test_util.clj
@@ -9,7 +9,8 @@
    [taoensso.timbre :as log])
   (:import
    [java.nio ByteBuffer]
-   [java.util Arrays Locale]))
+   [java.util Arrays Locale]
+   [java.util.concurrent CountDownLatch]))
 
 (set! *warn-on-reflection* true)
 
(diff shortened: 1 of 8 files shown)


modules/test-util/src/blaze/test_util.clj
```diff
@@ -88,6 +89,21 @@
    :timezone :utc}
   :output-fn output-fn})
 
+(defn submit-blocking-task!
```
alexanderkiel, line 92: Should return a release function so that the taks doesn't run forever. Tasks should be released after the timeout happend. Otherwise each test run will consume one thread.

alexanderkiel (changes requested):

alexanderkiel (approved):

PR #3956, opened 2026-07-27
Improve Documentation of blaze.page-id-cipher

### LINKED ISSUE #3951: Improve Documentation of blaze.page-id-cipher
The namespace `blaze.page-id-cipher` has no namespace docstring and most of its functions are undocumented. It is not obvious from the code:

* what the cipher is used for (encrypting the query params of paging links into opaque page IDs),
* why the Tink key set is stored as a `DocumentReference` resource in the **admin database** and how that resource is laid out (`identifier`, `content[0].attachment.data`),
* how the resource is created race-free across multiple Blaze instances (conditional create on the `identifier` clause),
* how key rotation is propagated to all instances (subscription to the changed-resources publisher of the admin node) and guarded against lost updates (`:if-match` on the version id),
* how long a page ID stays decryptable after a rotation.

Add a namespace docstring and function docstrings covering the above. Documentation only — no behaviour change.

alexanderkiel:
Closes: #3951

### DIFF
--- a/modules/page-id-cipher/src/blaze/page_id_cipher.clj
+++ b/modules/page-id-cipher/src/blaze/page_id_cipher.clj
@@ -1,4 +1,83 @@
 (ns blaze.page-id-cipher
+  "Cipher used to encrypt and decrypt page IDs.
+
+  Paging links Blaze hands out don't carry the query params of the page in the
+  clear. Instead the query params are encrypted into an opaque page ID that is
+  used as a path segment of the paging URL. This component provides the AEAD
+  primitive used for that. The encryption and decryption themselves happen in
+  `blaze.middleware.fhir.decrypt-page-id`.
+
+  Encrypting page IDs has three effects. First, clients can't tamper with the
+  query params of a page, because a page ID that wasn't produced by Blaze itself
+  won't decrypt. Second, internals like the point in time `t` of the database
+  value the page is based on, or the offset into the result set, stay opaque.
+  Third, page IDs expire, because the keys they were encrypted with are rotated
+  out of the key set; see the section on key rotation below.
+
+  ## Key Set Storage
+
+  The cipher is backed by a Tink key set. That key set has to be shared by all
+  Blaze instances in a deployment, because a paging link created by one instance
+  has to be resolvable by every other instance, and it has to survive restarts.
+  Rather than introducing a store of its own, the cipher keeps the key set in
+  the admin database (`:blaze.db.admin/node`) as a `DocumentReference` resource
+  with the following properties:
+
+  * `identifier` - a single identifier with the value `page-id-cipher` under
+    which the resource is looked up,
+  * `status` - `current`, because FHIR requires the property,
+  * `content[0].attachment.data` - the serialized key set, Base64 encoded.
+
+  A `DocumentReference` is used because it can carry an arbitrary binary
+  payload together with metadata. The key set goes into the base64Binary `data`
+  property of its first attachment and the well-known `identifier` value makes
+  the resource findable with the `identifier` search param, without having to
+  remember its logical ID anywhere.
+
+  The admin database is used because the key set is operational state of the
+  server, like the `Task` resources of the job scheduler. It doesn't belong in
+  the main database that is served by the FHIR API.
+
+  The key set itself is stored unencrypted, but the `DocumentReference` resource
+  holding it isn't accessible via any API. The FHIR API only ever reads from the
+  main database, and of the admin database, only `Task` resources are exposed.
+
+  ## Lifecycle
+
+  On init, the key set resource is looked up by its identifier and created with
+  a freshly generated key set if it doesn't exist yet. That create is
+  conditional on the very same identifier, so multiple Blaze instances starting
+  concurrently end up with exactly one key set resource. Transactions are
+  serialized through the transaction log, and the conditional creates of the
+  instances that lose the race are no-ops.
+
+  Each instance subscribes to the changed `DocumentReference` resources of the
+  admin node and refreshes its in-memory key set as soon as the key set resource
+  changes. That way a rotation performed by one instance reaches all other
+  instances.
+
+  ## Key Rotation
+
+  Paging links have to expire, and rotating the keys is what makes them expire.
+
+  A page ID carries the point in time `t` of the database value the paging
+  started at, and every following page is read from that same database value. So
+  a paging link keeps a fixed, historic database state reachable, including the
+  resources that were deleted after `t`. Clients must not be able to use such a
+  link indefinitely, because deleted resources have to become inaccessible
+  eventually, for GDPR reasons. Once the keys a page ID was encrypted with are
+  gone from the key set, the page ID no longer decrypts and the database state it
+  pointed to can't be reached through it anymore.
+
+  The keys are therefore rotated every `key-rotation-period`, one hour by
+  default, by writing an updated key set back into the key set resource. That
+  write uses the version ID of the resource read as `:if-match` precondition, so
+  two instances rotating at the same time can't lose each other's keys.
+
+  Rotation retires keys gradually (see `blaze.page-id-cipher.impl`): a key is
+  kept for three more rotation periods after it stopped being primary. So with
+  the default rotation period, a paging link stays usable for at least three
+  hours and is guaranteed to be dead after five."
   (:require
    [blaze.async.comp :as ac]
    [blaze.async.flow :as flow]
@@ -22,12 +101,16 @@
 
 (set! *warn-on-reflection* true)
 
-(def ^:private ^:const identifier
+(def ^:private ^:const key-set-identifier
+  "The identifier value the key set `DocumentReference` is looked up with."
   "page-id-cipher")
 
-(defn- find-key-set-resource [db]
+(defn- find-key-set-resource
+  "Returns a CompletableFuture that will complete with the resource handle of
+  the key set `DocumentReference` in `db`, or `nil` if it doesn't exist yet."
+  [db]
   (log/trace "try to find the key set resource")
-  (-> (d/type-query db "DocumentReference" [["identifier" identifier]])
+  (-> (d/type-query db "DocumentReference" [["identifier" key-set-identifier]])
       (ac/then-apply coll/first)))
 
 (defn- b64-encode [bytes]
@@ -45,20 +128,31 @@
   {:fhir/type :fhir.DocumentReference/content
    :attachment (key-set-attachment key-set-handle)})
 
-(defn- key-set-resource [context]
+(defn- key-set-resource
+  "Returns a new key set `DocumentReference` resource with a freshly generated
+  key set as first attachment."
+  [context]
   {:fhir/type :fhir/DocumentReference
    :id (m/luid context)
-   :identifier [(type/identifier {:value (type/string identifier)})]
+   :identifier [(type/identifier {:value (type/string key-set-identifier)})]
    :status #fhir/code "current"
    :content
    [(key-set-content (impl/gen-new-key-set-handle))]})
 
-(defn- key-set-resource-create-op [context]
-  [:create (key-set-resource context) [["identifier" identifier]]])
+(defn- key-set-resource-create-op
+  "Returns a transaction operator creating a new key set resource, conditional
+  on the absence of a resource with the key set identifier.
+
+  Because transactions are serialized through the transaction log, at most one
+  of the creates submitted by concurrently starting Blaze instances will take
+  effect. The others become no-ops."
+  [context]
+  [:create (key-set-resource context) [["identifier" key-set-identifier]]])
 
 (defn- find-or-create-key-set-resource
-  "Returns a CompletableFuture that will complete with the DocumentReference
-  resource with the key set as first attachment."
+  "Returns a CompletableFuture that will complete with the `DocumentReference`
+  resource with the key set as first attachment, creating that resource with a
+  freshly generated key set if it doesn't exist yet."
   {:arglists '([context])}
   ([{:keys [node] :as context}]
    (find-or-create-key-set-resource context (d/db node)))
@@ -79,16 +173,24 @@
       (TinkProtoKeysetFormat/parseKeyset (InsecureSecretKeyAccess/get))))
 
 (defn- decode-key-set-handle
-  {:argLists '([key-set-resource])}
+  "Returns the key set handle decoded from the first attachment of
+  `key-set-resource`."
+  {:arglists '([key-set-resource])}
   [{[{{:keys [data]} :attachment}] :content}]
   (parse-key-set (:value data)))
 
-(defn- decode-state [key-set-resource]
+(defn- decode-state
+  "Returns the cipher state decoded from `key-set-resource`: the key set handle
+  together with the AEAD primitive derived from it."
+  [key-set-resource]
   (let [key-set-handle (decode-key-set-handle key-set-resource)]
     {:key-set-handle key-set-handle
      :aead (impl/get-aead key-set-handle)}))
 
-(deftype TaskSubscriber [node state ^:volatile-mutable subscription]
+;; Subscriber to the changed `DocumentReference` resources of the admin node.
+;; Keeps `state` in sync with the key set resource, which is how rotations
+;; performed by one Blaze instance reach all other instances.
+(deftype DocumentReferenceSubscriber [node state ^:volatile-mutable subscription]
   Flow$Subscriber
   (onSubscribe [_ s]
     (set! subscription s)
@@ -97,8 +199,8 @@
     (log/trace "Got" (count document-reference-handles)
                "changed document-reference(s)")
     (run!
-     (fn [{[{{value :value} :value}] :identifier :as document-reference}]
-       (when (= identifier value)
+     (fn [{[identifier] :identifier :as document-reference}]
+       (when (= key-set-identifier (:value (:value identifier)))
          (log/debug "Refresh key set")
          (reset! state (decode-state document-reference))))
      @(d/pull-many node document-reference-handles))
@@ -108,6 +210,10 @@
     (flow/cancel! subscription))
   (onComplete [_]))
 
+;; The component itself. It's an AEAD primitive that delegates to the primitive
+;; of the current `state`, so that all users of the component pick up key
+;; rotations without having to be notified. The `future` is the handle of the
+;; scheduled key rotation, used to cancel it on shutdown.
 (defrecord Cipher [state future]
   Aead
   (encrypt [_ plaintext associatedData]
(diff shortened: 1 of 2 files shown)


modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
 (ns blaze.page-id-cipher
+  "Cipher used to encrypt and decrypt page ID's.
```
trobanga: AEAD primitive for encrypting and decrypting page IDs.

modules/page-id-cipher/src/blaze/page_id_cipher/impl.clj
```diff
@@ -13,7 +18,9 @@
 (def ^:private ^Parameters parameters
   PredefinedAeadParameters/AES128_GCM)
 
-(defn gen-new-key-set-handle []
+(defn gen-new-key-set-handle
+  "Returns a new key set handle with a single, primary key."
```
trobanga: The comma makes these coordinate adjectives (i.e. "a single and primary key"), which is not the meaning — there is one key, and it is primary.

```suggestion
  "Returns a new key set handle with a single primary key."
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
...
+
+  Paging links Blaze hands out don't carry the query params of the page in the
+  clear. Instead the query params are encrypted into an opaque page ID that is
+  used as path segment of the paging URL. This component provides the AEAD
+  primitive used for that. The encryption and decryption itself happens in
```
trobanga: Two grammar fixes: `as path segment` is missing its article, and `The encryption and decryption` is a compound subject, so it takes `themselves happen` rather than `itself happens`.

```suggestion
  clear. Instead the query params are encrypted into an opaque page ID that is
  used as a path segment of the paging URL. This component provides the AEAD
  primitive used for that. Encryption and decryption themselves happen in
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
...
+  `blaze.middleware.fhir.decrypt-page-id`.
+
+  Encrypting page ID's has three effects. First, clients can't tamper with the
+  query params of a page, because a page ID that wasn't produced by Blaze itself
+  won't decrypt. Second, internals like the point in time `t` of the database
+  value the page is based on, or the offset into the result set, stay opaque.
+  Third, page ID's expire, because the keys they were encrypted with are rotated
```
trobanga: `ID's` is a possessive, not a plural — the plural of an initialism takes a bare `s`: `IDs`. Also on line 2 and in `impl.clj:6`.

There is precedent for the apostrophe form in the repo (`modules/luid/README.md` writes `UUID's`/`ULID's`, `docs/implementation/cql.md:89` writes `ID's`), but it isn't consistent: `modules/luid/src/blaze/luid.clj:2` already uses `LUIDs`.

```suggestion
  Encrypting page IDs has three effects. First, clients can't tamper with the
  query params of a page, because a page ID that wasn't produced by Blaze itself
  won't decrypt. Second, internals like the point in time `t` of the database
  value the page is based on, or the offset into the result set, stay opaque.
  Third, page IDs expire, because the keys they were encrypted with are rotated
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
...
+  out of the key set. That last one is the reason the keys are rotated at all,
+  see the section on key rotation below.
```
trobanga: Comma splice — two independent clauses joined by a comma. A semicolon fixes it. Same pattern in `impl.clj:69-70` (`through a paging link, see ...`).

```suggestion
  out of the key set. That last one is the reason the keys are rotated at all;
  see the section on key rotation below.
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
...
+
+  ## Key Set Storage
+
+  The cipher is backed by a Tink key set. That key set has to be shared by all
+  Blaze instances of a deployment, because a paging link created by one instance
+  has to be resolvable by every other instance, and it has to survive restarts.
+  Rather than introducing a store of its own, the key set is kept in the admin
+  database (`:blaze.db.admin/node`) as a `DocumentReference` resource with the
+  following properties:
```
trobanga: Two things:

- `instances of a deployment` → `instances in a deployment`.
- `Rather than introducing a store of its own, the key set is kept ...` is a dangling modifier: the implied subject of *introducing* is *the key set*, which can't introduce a store. Making the cipher the subject fixes it.

```suggestion
  The cipher is backed by a Tink key set. That key set has to be shared by all
  Blaze instances in a deployment, because a paging link created by one instance
  has to be resolvable by every other instance, and it has to survive restarts.
  Rather than introducing a store of its own, the cipher keeps the key set in
  the admin database (`:blaze.db.admin/node`) as a `DocumentReference` resource
  with the following properties:
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
...
+
+  * `identifier` - a single identifier with the value `page-id-cipher` under
+    which the resource is looked up,
+  * `status` - `current`, because FHIR requires the property,
+  * `content[0].attachment.data` - the serialized key set, Base64 encoded.
+
+  A `DocumentReference` is used because it is the FHIR resource type meant to
+  carry an arbitrary binary payload together with metadata. The key set goes
```
trobanga: This overstates it slightly. The FHIR resource *meant* to carry an arbitrary binary payload is `Binary`; `DocumentReference` is the metadata wrapper that points at content — and can inline it in `content.attachment.data`, which is what happens here. The `together with metadata` part is what actually justifies the choice, so I'd weaken the claim:

> A `DocumentReference` is used because it can carry an arbitrary binary payload together with metadata.

Your call — this is a judgement about FHIR intent rather than a clear error.

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
...
+  into the base64Binary `data` property of its first attachment and the
+  well-known `identifier` value makes the resource findable with the
+  `identifier` search param, without having to remember its logical ID
+  anywhere.
+
+  The admin database is used because the key set is operational state of the
+  server, like the `Task` resources of the job scheduler. It doesn't belong into
```
trobanga: `belong into` is a Germanism (*gehört in*) — English takes `belong in`.

```suggestion
  server, like the `Task` resources of the job scheduler. It doesn't belong in
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
...
+  holding it isn't accessible via any API. The FHIR API only ever reads from the
+  main database, and of the admin database, only `Task` resources are exposed.
+
+  ## Lifecycle
+
+  On init, the key set resource is looked up by its identifier and created with
+  a freshly generated key set if it doesn't exist yet. That create is
+  conditional on the very same identifier, so multiple Blaze instances starting
+  concurrently end up with exactly one key set resource. Transactions are
+  serialized through the transaction log and the conditional creates of the
```
trobanga: `Transactions are serialized through the transaction log` and `the conditional creates ... are no-ops` are two independent clauses, so the `and` needs a comma before it.

```suggestion
  serialized through the transaction log, and the conditional creates of the
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
...
+  instances that lose the race are no-ops.
+
+  Each instance subscribes to the changed resources of type `DocumentReference`
+  of the admin node and refreshes its in-memory key set as soon as the key set
```
trobanga: `the changed resources of type X of the admin node` chains two `of`s and is hard to parse.

```suggestion
  Each instance subscribes to changes of `DocumentReference` resources in the
  admin node and refreshes its in-memory key set as soon as the key set
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
...
+  resource changes. That way a rotation performed by one instance reaches all
+  other instances.
+
+  ## Key Rotation
+
+  Paging links have to expire, and rotating the keys is what makes them expire.
```
trobanga, line 61: This restates what lines 14-16 already said (`Third, page IDs expire, because the keys ... are rotated out` / `That last one is the reason the keys are rotated at all`) — three statements of the same point across two sections. `expire` also appears twice in the one sentence.

I'd drop this line and let the next paragraph carry the section, or drop the forward reference on lines 15-16 and keep this one.

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
...
+  resources that were deleted after `t`. Clients must not be able to use such a
+  link indefinitely, because deleted resources have to become inaccessible
+  eventually, for GDPR reasons. Once the keys a page ID was encrypted with are
+  gone from the key set, the page ID no longer decrypts and the database state it
+  pointed to can't be reached through it anymore.
+
+  The keys are therefore rotated every `key-rotation-period`, one hour by
+  default, by writing an updated key set back into the key set resource. That
+  write uses the version ID of the resource read as `:if-match` precondition, so
+  two instances rotating at the same time can't lose each others keys.
```
trobanga: `each others` → `each other's`. An apostrophe missing where one actually belongs.

```suggestion
  two instances rotating at the same time can't lose each other's keys.
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -1,4 +1,86 @@
...
+
+  Rotation retires keys gradually (see `blaze.page-id-cipher.impl`), keeping the
+  keys a page ID could have been encrypted with around for three more rotation
+  periods after they stopped being the primary key. So with the default rotation
+  period, a paging link stays usable for at least three hours and is guaranteed
+  to be dead after five."
```
trobanga: `keeping the keys ... with around for three more rotation periods` is a garden path — the `around` belongs to `keeping ... around`, seven words back — and `they stopped being the primary key` disagrees in number.

The arithmetic is right, incidentally: I traced a key through `rotate-keys`, and both the "three more rotation periods" and the three-to-five hour window check out.

```suggestion
  Rotation retires keys gradually (see `blaze.page-id-cipher.impl`): a key is
  kept for three more rotation periods after it stopped being primary. So with
  the default rotation period, a paging link stays usable for at least three
  hours and is guaranteed to be dead after five."
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -22,12 +104,17 @@
 
 (set! *warn-on-reflection* true)
 
-(def ^:private ^:const identifier
+(def ^:private ^:const key-set-identifier
+  "The value of the identifier of the `DocumentReference` resource holding the
+  key set."
```
trobanga: `The value of the identifier of the ...` chains two `of`s for what is one idea.

```suggestion
  "The identifier value the key set `DocumentReference` is looked up with."
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -22,12 +104,17 @@
...
   "page-id-cipher")
 
-(defn- find-key-set-resource [db]
+(defn- find-key-set-resource
+  "Returns a CompletableFuture that will complete with the resource handle of
+  the key set `DocumentReference` in `db` or nil if it doesn't exist yet."
```
trobanga: Needs a comma before `or nil`, and `nil` should be backticked to match the `t` / `db` markup used elsewhere in these docstrings.

```suggestion
  the key set `DocumentReference` in `db`, or `nil` if it doesn't exist yet."
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -79,16 +177,24 @@
...
+  "Returns the key set handle decoded from the first attachment of
+  `key-set-resource`."
+  {:arglists '([key-set-resource])}
   [{[{{:keys [data]} :attachment}] :content}]
   (parse-key-set (:value data)))
 
-(defn- decode-state [key-set-resource]
+(defn- decode-state
+  "Returns the state of the cipher, the key set handle together with the AEAD
+  primitive derived from it, decoded from `key-set-resource`."
```
trobanga: Three appositives stacked on each other (`the state of the cipher` / `the key set handle together with ...` / `decoded from ...`), so the sentence stays open to the very end. Moving the source up front unpacks it:

```suggestion
  "Returns the cipher state decoded from `key-set-resource`: the key set handle
  together with the AEAD primitive derived from it."
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -79,16 +177,24 @@
...
+  [key-set-resource]
   (let [key-set-handle (decode-key-set-handle key-set-resource)]
     {:key-set-handle key-set-handle
      :aead (impl/get-aead key-set-handle)}))
 
-(deftype TaskSubscriber [node state ^:volatile-mutable subscription]
+;; Subscriber on the changed `DocumentReference` resources of the admin node
+;; that keeps `state` in sync with the key set resource. It's the way rotations
+;; performed by one Blaze instance reach all other instances.
```
trobanga: `Subscriber on` → `Subscriber to`. Also, `resources of the admin node that keeps `state` in sync` reads as though the *node* is what keeps state in sync; splitting the sentence removes the ambiguity.

```suggestion
;; Subscriber to the changed `DocumentReference` resources of the admin node.
;; Keeps `state` in sync with the key set resource, which is how rotations
;; performed by one Blaze instance reach all other instances.
```

modules/page-id-cipher/src/blaze/page_id_cipher.clj
```diff
@@ -108,37 +214,59 @@
...
+
 (defn- rotate-keys
-  "Rotates the keys in the key set available at DocumentReference with
-  identifier `page-id-cipher` in `node`."
+  "Rotates the keys in the key set of the `DocumentReference` resource with the
+  identifier `page-id-cipher` in `node`.
+
+  Writing the rotated key set back into the resource will notify all Blaze
+  instances subscribed to the changed `DocumentReference` resources of `node`,
+  including this one."
```
trobanga: `this one` could refer to the instance or to the resource. Naming it removes the ambiguity.

```suggestion
  including this instance."
```

modules/page-id-cipher/src/blaze/page_id_cipher/impl.clj
```diff
@@ -1,4 +1,9 @@
 (ns blaze.page-id-cipher.impl
+  "Tink key set handling of the page ID cipher.
```
trobanga: `handling of the page ID cipher` reads as a possessive, as though the cipher owns the handling. `for` states the relationship you mean.

```suggestion
  "Tink key set handling for the page ID cipher.
```

modules/page-id-cipher/src/blaze/page_id_cipher/impl.clj
```diff
@@ -1,4 +1,9 @@
...
+
+  Key sets consist of AES128-GCM keys. Exactly one key of a key set is the
+  primary key that is used for encryption. All keys of a key set can be used for
+  decryption, which is what makes page ID's outlive a key rotation."
```
trobanga: See the note on `page_id_cipher.clj:10` — plural initialisms take a bare `s`.

```suggestion
  decryption, which is what makes page IDs outlive a key rotation."
```

modules/page-id-cipher/src/blaze/page_id_cipher/impl.clj
```diff
@@ -53,15 +60,26 @@
   [old-key primary-key] -> [old-key primary-key new-key]
   [old-key primary-key new-key] -> [old-key old-key primary-key]
   [old-key old-key primary-key] -> [old-key primary-key new-key]
-  [old-key primary-key new-key] -> [old-key old-key primary-key]"
+  [old-key primary-key new-key] -> [old-key old-key primary-key]
+
+  So a new key is first added as non-primary key, is made primary on the next
```
trobanga: Missing article: `as a non-primary key`.

```suggestion
  So a new key is first added as a non-primary key, is made primary on the next
```

modules/page-id-cipher/src/blaze/page_id_cipher/impl.clj
```diff
@@ -53,15 +60,26 @@
...
+  rotation, stays primary for two rotations and is only removed three rotations
+  after it stopped being primary. That way a page ID stays decryptable for at
+  least three and at most five rotation periods after it was created. The upper
+  bound is what bounds the time a historic database state stays reachable
+  through a paging link, see `blaze.page-id-cipher`."
```
trobanga: `The upper bound is what bounds` repeats the word, and `through a paging link, see ...` is the same comma splice as `page_id_cipher.clj:15`.

```suggestion
  least three and at most five rotation periods after it was created. That
  upper bound limits how long a historic database state stays reachable through
  a paging link; see `blaze.page-id-cipher`."
```

trobanga (changes requested):

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: AEAD is mentioned later, but I take ID's -> IDs

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: I weakened the statement.

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher/impl.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher/impl.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher/impl.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher/impl.clj
alexanderkiel: done

modules/page-id-cipher/src/blaze/page_id_cipher/impl.clj
alexanderkiel: done

trobanga (approved):

PR #3968, opened 2026-07-28
Add Integration Test for Validator Modes

### LINKED ISSUE #3847: Add Synthea Integration Test for External Validator Failure Modes
Blaze supports external profile validation with three failure modes (tag-only, tag-outcome, reject), but the integration tests only exercise tag-outcome with MII/KDS data. The other two modes and the Synthea dataset against US Core profiles are untested.

There should be an integration test that loads Synthea data through Blaze with the external validator enabled and covers all three failure modes, so regressions in tagging and rejection behavior are caught in CI.

knoppiks:


knoppiks:
#3998 found while implementing. Rebase after #4000 lands

.github/integration-test-kds/docker-compose.yml
```diff
@@ -53,6 +53,17 @@ services:
     - "8084:8080"
     volumes:
     - ./fhir-settings.json:/app/fhir-settings.json:ro
```
alexanderkiel: Not needed anymore, because the mii-fhir-validator generated the fhir-settings.json at startup. Please remove this volume mount and the file.

.github/integration-test-extern-validator/check-reject-mode.sh
```diff
@@ -0,0 +1,96 @@
...
+
+bundle > "$request_body"
+
+test "transaction status of the invalid Patient" "$(send POST "$base")" "400"
+test_rejection_outcome "transaction"
+
+# --- no invalid resource was persisted --------------------------------------
+
+test "number of persisted invalid Patients" \
+  "$(curl -s -H 'Accept: application/fhir+json' "$base/Patient?_tag=${tag_system}|invalid&_summary=count" | jq -r '.total')" "0"
```
alexanderkiel: Please use search_strict here

.github/integration-test-extern-validator/create-synthea-subset.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -euo pipefail
+
+# Creates a subset of the Synthea test data in the `synthea-subset` directory.
+# Validating every resource of the full Synthea data set against the external
+# validator would be too slow, so only three of the 121 patient bundles are
```
alexanderkiel: this are not three patient bundles, instead you have two master data bundles and one patient bundle

alexanderkiel (changes requested):
Please load USCore into the terminology server and active SNOMED CT.

PR #3979, opened 2026-07-28
Replace Deprecated Tink Keyset Format Methods

### LINKED ISSUE #3966: Replace Deprecated Tink Keyset Format Methods
`blaze.page-id-cipher` calls two `TinkProtoKeysetFormat` methods that are deprecated in the Tink version we depend on (1.23.0):

* `modules/page-id-cipher/src/blaze/page_id_cipher.clj:121` — `TinkProtoKeysetFormat/serializeKeyset`
* `modules/page-id-cipher/src/blaze/page_id_cipher.clj:173` — `TinkProtoKeysetFormat/parseKeyset`

Verified against `tink-1.23.0.jar` with `javap -v`. Only the two-argument overloads carry `Deprecated: true`; the three-argument overloads taking a `Configuration` do not:

```
parseKeyset(byte[], SecretKeyAccess)                    Deprecated: true
parseKeyset(byte[], SecretKeyAccess, Configuration)     -
serializeKeyset(KeysetHandle, SecretKeyAccess)          Deprecated: true
serializeKeyset(KeysetHandle, SecretKeyAccess, Configuration)  -
```

## Proposed change

Pass a `Configuration` as the third argument. `RegistryConfiguration/get` is already the configuration used everywhere else in the code base, so it is the natural choice:

* `modules/page-id-cipher/src/blaze/page_id_cipher/impl.clj:83`
* `modules/fhir-test-util/src/blaze/fhir/test_util.clj:102`
* `modules/rest-util/test/blaze/middleware/fhir/decrypt_page_id_test.clj:26`

`RegistryConfiguration` has to be added to the `:import` of `blaze.page-id-cipher`.

## Acceptance criteria

* No deprecated Tink methods are called any more.
* Key set resources written by an older Blaze version still parse after the change, so that a running deployment keeps its key set across the upgrade. The `Configuration` argument selects how primitives are created rather than how the keyset is encoded, so the serialized form is expected to be unchanged — this should be confirmed by a test that parses a keyset serialized with the current code.

## Note

Clojure interop does not warn on deprecated Java methods, and `clj-kondo` does not check for them either, so neither `make lint` nor compilation flags these calls. Other uses of deprecated Java APIs would be equally invisible; this one was found by reading the Tink API while reviewing #3956.

alexanderkiel:
Closes: #3966

### DIFF
--- a/modules/page-id-cipher/src/blaze/page_id_cipher.clj
+++ b/modules/page-id-cipher/src/blaze/page_id_cipher.clj
@@ -95,7 +95,7 @@
    [java-time.api :as time]
    [taoensso.timbre :as log])
   (:import
-   [com.google.crypto.tink Aead InsecureSecretKeyAccess TinkProtoKeysetFormat]
+   [com.google.crypto.tink Aead]
    [java.util Base64]
    [java.util.concurrent Flow$Subscriber]))
 
@@ -118,7 +118,7 @@
 
 (defn- encode-key-set-handle [key-set-handle]
   (-> key-set-handle
-      (TinkProtoKeysetFormat/serializeKeyset (InsecureSecretKeyAccess/get))
+      (impl/serialize-key-set)
       (b64-encode)))
 
 (defn- key-set-attachment [key-set-handle]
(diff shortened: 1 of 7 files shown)


trobanga (changes requested):
you should add function specs following the codebase's convention

trobanga (approved):

PR #3991, opened 2026-07-29
Implement Multiply for Integer

### LINKED ISSUE #113: CQL: Implement Multiply for Integer
No implementation of method: :multiply of protocol: #'blaze.elm.protocols/Multiply found for class: java.lang.Integer

trobanga:
Closes #113

### DIFF
--- a/modules/cql/test/blaze/elm/integer_test.clj
+++ b/modules/cql/test/blaze/elm/integer_test.clj
@@ -1,0 +1,16 @@
+(ns blaze.elm.integer-test
+  (:require
+   [blaze.elm.integer]
+   [blaze.elm.protocols :as p]
+   [blaze.test-util :as tu]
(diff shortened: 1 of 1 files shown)


modules/cql/test/blaze/elm/compiler/arithmetic_operators_test.clj
```diff
@@ -770,6 +770,9 @@
       "1" "2" 2
       "2" "2" 4)
 
+    (testing "java.lang.Integer instances"
```
alexanderkiel: Please move that test into a new test namespace `blaze.elm.integer-test`. It doesn't belong here at the compiler level.

alexanderkiel (changes requested):

alexanderkiel (approved):

PR #3993, opened 2026-07-29
Fix Protocol Warnings in Module CQL

### LINKED ISSUE #2567: Fix Protocol Warnings in Module CQL
clj-kondo is complaining about missing protocol methods in the CQL module. We should fix them.

trobanga:
Closes #2567


trobanga:
We don't want to test `equal` and `hashCode` here.

PR #3996, opened 2026-07-29
Fix Flashing Loading Indicator on Search Pages

### LINKED ISSUE #3994: Fix Flashing Loading Indicator on Search Pages
Running a second search on a resource type page makes the "loading..." text appear immediately, even when the results come back straight away. It flickers in and out again.

The indicator is meant to stay hidden for fast searches and only appear once a search has been running for a noticeable moment.

There is a second part to it. While the indicator is flashing, the elapsed time shown next to it is not the time the current search has been running. It starts at however long ago the *previous* search was started — often many seconds or minutes if the page has been sitting open — and then drops back down.

Steps to reproduce:

1. Open a resource type page and run a search that returns quickly.
2. Leave the page open for a minute or so.
3. Run another search from the search form.

Expected: no loading indicator at all, because the search is fast.

Observed: "loading..." appears for a moment, showing an elapsed time of about a minute, then disappears.

The same thing happens when clicking further through the pages of a large result set: the first page change is fine, but each one after that briefly shows the indicator with an elapsed time carried over from the page before.

The very first search after opening a page is also fine — the indicator only misbehaves from the second search onwards.

knoppiks:
Closes: #3994

### DIFF
--- a/modules/frontend/src/lib/loading-indicator.svelte
+++ b/modules/frontend/src/lib/loading-indicator.svelte
@@ -1,0 +1,43 @@
+<script lang="ts">
+  import { fade, slide } from 'svelte/transition';
+
+  interface Props {
+    start: number;
+  }
+
+  let { start }: Props = $props();
+
+  const delayMillis = 300;
+
+  let now = $state(Date.now());
+
+  $effect(() => {
+    const interval = setInterval(() => (now = Date.now()), 100);
+    return () => clearInterval(interval);
+  });
+
+  /**
+   * Deriving the elapsed time from the clock and the current `start` — instead
+   * of accumulating it into state — makes it drop back to zero the moment a new
+   * query starts. The clock lags up to one tick behind, so a `start` taken after
+   * the last tick is clamped to zero.
+   */
+  let elapsed = $derived(Math.max(0, now - start));
+</script>
+
+{#if elapsed > delayMillis}
+  <div
+    in:fade|global={{ duration: 200 }}
+    out:slide|global={{ duration: 200 }}
+    class="text-center px-4 py-5 sm:px-6 text-gray-700 dark:text-gray-300"
+  >
+    <code>
+      loading...
+      {(elapsed / 1000).toLocaleString(undefined, {
+        minimumFractionDigits: 1,
+        maximumFractionDigits: 1
+      })}
+      s
+    </code>
+  </div>
+{/if}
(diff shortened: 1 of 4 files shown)


modules/frontend/src/lib/elapsed.svelte.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
alexanderkiel: the file should not be called `elapsed.svelte.ts`. Instead it should be just called `elapsed.ts`.

modules/frontend/src/lib/loading-indicator.svelte
```diff
@@ -0,0 +1,30 @@
+<script lang="ts">
+  import { elapsedMillis } from './elapsed.svelte.js';
```
alexanderkiel: Should be imported via `$lib/elapsed.js`. Not a relative path.

modules/frontend/src/lib/loading-indicator.svelte
```diff
@@ -0,0 +1,30 @@
...
+  import { fade, slide } from 'svelte/transition';
+
+  interface Props {
+    start: number;
+    delayMillis?: number;
+  }
+
+  let { start, delayMillis = 300 }: Props = $props();
+
+  const elapsed = elapsedMillis(() => start);
```
alexanderkiel: Why is `start` a function? Can't it be just a value?

modules/frontend/src/lib/elapsed.svelte.ts
alexanderkiel: But I would just put that function, if really needed, into the `loading-indicator` component.

modules/frontend/src/lib/loading-indicator.svelte
alexanderkiel: Ok, it is necessary, so that the `elapsedMillis` function gets a new start value on new data loaded. But this should be possible with a direct `$effect` here in the component. Search for `$effect` in other components for examples.

modules/frontend/eslint.config.js
```diff
@@ -19,7 +19,7 @@ export default ts.config(
     }
   },
   {
-    files: ['**/*.svelte'],
+    files: ['**/*.svelte', '**/*.svelte.js', '**/*.svelte.ts'],
```
alexanderkiel: There should be no `*.svelte.js/ts`.

alexanderkiel (changes requested):

alexanderkiel (approved):

PR #4009, opened 2026-07-31
Reduce Remaining java-time Allocations on Hot Paths

### LINKED ISSUE #3961: Reduce Remaining java-time Allocations on Hot Paths
Follow-up to #3959, which replaced the `instant` and `offset-date-time` clock arities with `blaze.time`. This issue covers the remaining `java-time.api` calls in production code.

Every `java-time.api` call in `modules/*/src` and `src` was inventoried and measured against its direct `java.time` equivalent with `ThreadMXBean.getThreadAllocatedBytes` (clojure.java-time 1.4.3, JDK 25, 500k iterations, result written to a volatile to prevent escape analysis from removing the allocation). Allocation figures are deterministic; timings were taken on a loaded machine and are indicative only.

## Findings

| call | bytes/call | direct `java.time` | factor |
|---|---:|---:|---:|
| `(time/duration odt1 odt2)` | 1332 | `Duration/between` — 32 | 41× |
| `(time/instant odt)` | 1292 | `.toInstant` — 33 | 40× |
| `(time/minus ld (months 1) (days 2))` | 160 | 48 | 3.3× |
| `(time/plus ld (months 1) (days 2))` | 160 | 31 | 5.2× |
| `(time/plus odt dur)` | 137 | `.plus` — 82 | 1.7× |
| `(time/as dur :millis)` | 121 | `.toMillis` — 30 | 4.0× |
| `(time/as dur :nanos)` | 93 | `.toNanos` — 32 | 2.9× |
| `(time/after? a b)` | 9.3 | `.isAfter` — 5.7 | 1.6× |
| `(time/before? a b)` | 9.6 | `.isBefore` — 5.7 | 1.7× |

Two distinct causes:

* `duration` and `instant` with a non-clock argument are **implicit `deffactory` arities**, so they go through `java-time.defconversion/call-conversion` — the same runtime conversion-graph resolution as #3959, at the same ~1.3 KiB per call.
* `plus`, `minus` and `as` are ordinary varargs functions (`[o & os]`), so each call allocates an `ArraySeq` for the rest args on top of protocol dispatch.

## What is already fine — no action needed

The amount constructors allocate no more than the `java.time` object they return, so there is nothing to gain:

| call | bytes/call | direct |
|---|---:|---:|
| `(time/months 1)` | 30.2 | `Period/ofMonths` — 29.9 |
| `(time/days 2)` | 30.2 | `Period/ofDays` — 29.9 |
| `(time/nanos 1000000)` | 29.7 | `Duration/ofNanos` — 30.2 |
| `(time/seconds 1)` | 30.2 | `Duration/ofSeconds` — 30.2 |
| `(time/hours 1)` | 30.2 | — |
| `(time/millis 10)` | 33.3 | `Duration/ofMillis` — 30.2 |

All remaining uses are `ig/init-key` defaults, `m/pre-init-spec` predicates (`time/duration?`, `time/clock?`), or `blaze.fhir.test-util` clock construction — evaluated once at startup or in tests, so they are irrelevant regardless of cost.

## Call sites worth changing, by impact

**1. CQL date arithmetic — scales with population size.** `modules/cql/src/blaze/elm/date_time.clj` lines 420, 427, 434 (`add`) and 515, 526 (`subtract`) use `time/plus` / `time/minus`. These implement `p/add` / `p/subtract`, which the compiled ELM `Add` and `Subtract` operators call (`elm/compiler/arithmetic_operators.clj:21,146`) — once per resource per expression during measure evaluation. At ~130 bytes of avoidable garbage per call this is the only site that scales with the millions-of-patients workload the project targets.

Lines 473 and 580 use `time/after?` / `time/before?` on the same path, but the delta is under 4 bytes per call and probably not worth the churn.

**2. Local tx-log poll loop.** `modules/db/src/blaze/db/tx_log/local.clj:140` calls `(time/as timeout :millis)` on every waiting iteration of `poll!`. Replaceable with `(.toMillis ^Duration timeout)`.

**3. The two 1.3 KiB conversion arities.** Low call frequency but the worst per-call cost and the most trivial fixes:
   * `modules/job-disk-perf/src/blaze/job/disk_perf.clj:163` and `modules/job-re-index/src/blaze/job/re_index.clj:90` — `(time/duration a b)` → `Duration/between`
   * `modules/operation-measure-evaluate-measure/src/blaze/fhir/operation/evaluate_measure/measure.clj:426` — `(time/instant (time/plus now timeout))` → `.toInstant`, and the enclosing `time/plus` → `.plus`

**4. Scheduler.** `modules/scheduler/src/blaze/scheduler.clj:41-42` uses `time/as` with `:nanos`, but only once per scheduled task, so this is cosmetic.

`modules/db/src/blaze/db/spec.clj:38` passes `time/instant` to a generator and is test-only; it can stay.

## Suggested scope

Items 1–3. Item 1 is the only one that matters for throughput; items 2 and 3 are one-line changes that remove the last of the conversion-graph calls from production code.

trobanga:
Closes #3961

### DIFF
--- a/modules/util/src/blaze/time.clj
+++ b/modules/util/src/blaze/time.clj
@@ -1,12 +1,14 @@
 (ns blaze.time
-  "Functions for creating temporal entities.
+  "Functions for creating and calculating with temporal entities.
 
-  The `clojure.java-time` counterparts of these functions define only their
-  zero-argument arity explicitly. Every other arity is routed through a
-  conversion graph that is resolved at runtime, on every call, allocating about
-  2 KiB per call. The functions here delegate to `java.time` directly."
+  The `clojure.java-time` counterparts of these functions are either routed
+  through a conversion graph that is resolved at runtime, on every call,
+  allocating between 1.3 and 2 KiB per call, or are varargs functions that
+  allocate an `ArraySeq` for their rest args. The functions here delegate to
+  `java.time` directly."
   (:import
-   [java.time Clock Instant OffsetDateTime]))
+   [java.time Clock Duration Instant OffsetDateTime]
+   [java.time.temporal Temporal TemporalAmount TemporalUnit]))
 
 (set! *warn-on-reflection* true)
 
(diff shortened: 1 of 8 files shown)


modules/util/src/blaze/time.clj
```diff
@@ -20,3 +22,65 @@
...
+  "Returns `temporal` with `amount` units of `unit` added.
+
+  Returns `temporal` unchanged if `amount` is zero, in order to avoid the
+  allocation of a new temporal."
+  ^Temporal [^Temporal temporal amount ^TemporalUnit unit]
+  (if (zero? amount)
+    temporal
+    (.plus temporal (long amount) unit)))
+
+(defn plus-months-days
```
alexanderkiel: Such functions will not scale. Please use two calls to `plus-unit` instead.

alexanderkiel (changes requested):

alexanderkiel (approved):

PR #4013, opened 2026-07-31
Fix Parsing of Negative Time Zone Offsets

### LINKED ISSUE #4012: Fix Parsing of Negative Time Zone Offsets
Parsing a `dateTime` or `instant` with a negative time zone offset that has a non-zero minute part fails:

```clojure
(system/parse-date-time "2020-01-02T03:04:05-05:30")
;; => #:cognitect.anomalies{:category :incorrect
;;                          :message "Zone offset minutes and seconds must be negative because hours is negative"}
```

`DateTime.parseOffset` negates only the hour part:

```java
return ZoneOffset.ofHoursMinutes(c == '+' ? hour : -hour, minute);
```

`ZoneOffset` rejects a negative hour combined with a positive minute, so every negative offset with a non-zero minute part is refused. Positive offsets and offsets with a zero minute part are unaffected, which is why this went unnoticed.

Affected are the real-world offsets of India (`-05:30` for the western hemisphere equivalent), Newfoundland (`-03:30`), the Marquesas Islands (`-09:30`) and several others, whenever they appear in a `dateTime`, `instant` or any search parameter value parsed through `system.DateTime/parse`.

Both parts have to carry the sign:

```java
return c == '+'
        ? ZoneOffset.ofHoursMinutes(hour, minute)
        : ZoneOffset.ofHoursMinutes(-hour, -minute);
```

This is the only `ZoneOffset.ofHoursMinutes` call in production code.

## Why the property tests missed it

`blaze.fhir.spec.generators/zone-offset` only ever generated offsets with a zero minute part:

```clojure
(def zone-offset
  (gen/fmap (partial apply format "%s%02d:00")
            (gen/tuple (gen/elements ["+" "-"]) (gen/choose 1 14))))
```

So every property test over `dateTime` and `instant` values in `fhir-structure`, `cql` and `db` was blind to the whole class. Widening it to generate a minute part makes the suite fail with 39 errors and 4 failures against the unfixed code.

Note that FHIR only allows a non-zero minute part for the hours `00`-`13`, while `14` is restricted to `14:00`, so the generator has to keep `+14:15` and friends out.

alexanderkiel:
Closes: #4012

### DIFF
--- a/modules/fhir-structure/java/blaze/fhir/spec/type/system/DateTime.java
+++ b/modules/fhir-structure/java/blaze/fhir/spec/type/system/DateTime.java
@@ -192,7 +192,9 @@ public interface DateTime extends JavaSystemType, Temporal {
             int hour = parseInt(s, idx + 1, idx + 3);
             int minute = parseInt(s, idx + 4, idx + 6);
             if (s.length() == idx + 6) {
-                return ZoneOffset.ofHoursMinutes(c == '+' ? hour : -hour, minute);
+                return c == '+'
+                        ? ZoneOffset.ofHoursMinutes(hour, minute)
+                        : ZoneOffset.ofHoursMinutes(-hour, -minute);
             } else {
                 throw new DateTimeParseException("Text cannot be parsed to a DateTime.", s, idx + 6);
             }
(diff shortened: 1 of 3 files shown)


EmteZogaf (approved):
lgtm

PR #4028, opened 2026-08-04
Update FHIR Validator to v6.10.0

### LINKED ISSUE #4025: Update FHIR Validator to v6.10.0
The update of the FHIR Validator to [v6.10.0](https://github.com/hapifhir/org.hl7.fhir.core/releases/tag/6.10.0) fails in CI (`integration-test-validation` and `terminology-tests`) because v6.10.0 adds SSRF protection which refuses plain HTTP requests:

```
org.hl7.fhir.exceptions.TerminologyServiceException: org.hl7.fhir.exceptions.FHIRException:
  Error fetching the server's capability statement:
  Refusing to fetch from non-https URL: http://localhost:8080/fhir/metadata
```

Both the validation script and the terminology test scripts use the Blaze instance started in the job as terminology server via `-tx http://localhost:8080/fhir`.

The documented way to allow HTTP for a particular server is a `fhir-settings.json` with an `allowHttp` entry for that server, like the one we already have in `.github/integration-test-kds/fhir-settings.json`. However, that doesn't work for `validator_cli.jar` v6.10.0: the CLI only sets the settings file path via `FhirSettings.setExplicitFilePath`, but nothing ever calls `ManagedWebAccess.loadFromFHIRSettings`, so the `servers` list (and therefore `allowHttp`/`allowPrivateNetwork`) is never handed to the HTTP client. Verified locally with v6.10.0: the request is still refused with a `fhir-settings.json` in place, no matter whether it is passed via `-fhir-settings` or placed at `~/.fhir/fhir-settings.json`.

What does work is the `-ssrf-protection-enabled=false` command line option. That is acceptable here because both jobs only talk to the local Blaze instance and validate our own test data.

Tasks:

* update `VALIDATOR_VERSION`/`VALIDATOR_CHECKSUM` to v6.10.0
* pass `-ssrf-protection-enabled=false` in `.github/validation/validate.sh` and in the `.github/terminology-tests` scripts

alexanderkiel:
Closes: #4025

### DIFF
--- a/.github/terminology-tests/all.sh
+++ b/.github/terminology-tests/all.sh
@@ -2,6 +2,6 @@
 set -euo pipefail
 
 java -jar validator_cli.jar txTests \
-  -tx http://localhost:8080/fhir \
+  -tx http://localhost:8080/fhir -ssrf-protection-enabled=false \
   -output .github/terminology-tests/output \
   -mode flat
--- a/.github/terminology-tests/good.sh
+++ b/.github/terminology-tests/good.sh
@@ -2,7 +2,7 @@
 set -euo pipefail
 
 java -jar validator_cli.jar txTests -test-version 1.7.9 \
-  -tx http://localhost:8080/fhir \
+  -tx http://localhost:8080/fhir -ssrf-protection-enabled=false \
   -output .github/terminology-tests/output \
   -mode flat \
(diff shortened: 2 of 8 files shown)


EmteZogaf (approved):
lgtm

PR #4049, opened 2026-08-06
Prevent Invalid Anomalies Passing Through ba/anomaly

### LINKED ISSUE #4046: Prevent Invalid Anomalies Passing Through ba/anomaly
The function spec of `blaze.anomaly/anomaly` promises `:ret (s/nilable ::anom/anomaly)`:

```clojure
;; modules/anomaly/src/blaze/anomaly_spec.clj:72
(s/fdef ba/anomaly
  :args (s/cat :x any?)
  :ret (s/nilable ::anom/anomaly))
```

The `Map` branch of the `ToAnomaly` protocol breaks that promise:

```clojure
;; modules/anomaly/src/blaze/anomaly.clj:174
Map
(-anomaly [m]
  (when (anomaly? m)
    m))
```

`anomaly?` only tests whether the value is a map that contains the key `::anom/category`. It never inspects the value of that key, and it never inspects `::anom/message`. So the map is returned unchanged:

```clojure
(ba/anomaly {::anom/category ::bogus})
;; => {::anom/category ::bogus}                            ; not one of the 9 categories

(ba/anomaly {::anom/category nil})
;; => {::anom/category nil}

(ba/anomaly {::anom/category ::anom/fault ::anom/message :kw})
;; => {::anom/category ::anom/fault ::anom/message :kw}    ; message is not a string
```

This is the same class of problem that #4040 fixed for the `ExceptionInfo` branch. After that fix, the two branches of the same function give different guarantees.

One producer of a non-literal category exists:

```clojure
;; modules/db/src/blaze/db/impl/index/tx_error.clj:25
(cond-> {::anom/category (keyword "cognitect.anomalies" category)}
```

Here `category` is a string read back from the database. In normal operation Blaze wrote a valid one, so only damaged or very old stored data produces a bad category. The severity is therefore low. `st/instrument` checks `:args` only, not `:ret`, so no test fails today.

## Proposed fix

Keep `anomaly?` as it is. It is a cheap predicate used in hot paths. Making it strict would turn a map with a bad category into a *success* value downstream, which is worse than an invalid anomaly. Coerce instead of reject.

Extract a private helper and use it in both branches:

```clojure
(defn- sanitize
  "Replaces an invalid category with `::anom/fault` and removes an invalid
  message."
  [{::anom/keys [category message] :as m}]
  (cond-> m
    (nil? (categories category)) (assoc ::anom/category ::anom/fault)
    (and (some? message) (not (string? message))) (dissoc ::anom/message)))

Map
(-anomaly [m]
  (when (anomaly? m)
    (sanitize m)))
```

On the happy path this costs two map lookups, one hash-set lookup on an interned keyword and one `instance?` check. Both `cond->` tests are false, so `cond->` returns the input map itself — no allocation, no copy.

`ex-info-anomaly` can reuse `sanitize` for the category. Its message rule differs, because it falls back to the message of the exception.

## Acceptance

Write the test first. It must fail before the change:

```clojure
(is (s/valid? ::anom/anomaly (ba/anomaly {::anom/category ::bogus})))
```

Cover a bad category, a `nil` category and a non-string message.

trobanga:
Closes #4046

### DIFF
--- a/modules/anomaly/src/blaze/anomaly.clj
+++ b/modules/anomaly/src/blaze/anomaly.clj
@@ -151,6 +151,17 @@
   The spec of `::anom/category` is a set of all categories."
   (s/form ::anom/category))
 
+(defn- sanitize
+  "Replaces an invalid category of `m` with `::anom/fault` and removes an
+  invalid message.
+
+  Without that coercion, a map with an arbitrary category or a non-string
+  message would pass as an anomaly without satisfying `::anom/anomaly`."
+  [{::anom/keys [category message] :as m}]
+  (cond-> m
+    (not (categories category)) (assoc ::anom/category ::anom/fault)
+    (not (string? message)) (dissoc ::anom/message)))
+
 (defn- ex-info-anomaly
   "Builds an anomaly out of the message `msg` and the data `data` of an
   `ExceptionInfo`.
@@ -158,12 +169,9 @@
   The category and message of `data` are only used if they are valid. Otherwise
   the data of an arbitrary exception could turn the result into something that
   isn't an anomaly anymore."
-  [msg data]
-  (let [{::anom/keys [category message]} data
-        message (if (string? message) message msg)]
-    (cond-> (assoc data ::anom/category (if (categories category) category ::anom/fault))
-      (nil? message) (dissoc ::anom/message)
-      message (assoc ::anom/message message))))
+  [msg {::anom/keys [message] :as data}]
+  (cond-> (sanitize data)
+    (and (some? msg) (not (string? message))) (assoc ::anom/message msg)))
 
 (defprotocol ToAnomaly
   (-anomaly [x]))
@@ -192,7 +200,7 @@
   Map
   (-anomaly [m]
     (when (anomaly? m)
-      m))
+      (sanitize m)))
   Object
   (-anomaly [_])
   nil
--- a/modules/anomaly/test/blaze/anomaly_test.clj
+++ b/modules/anomaly/test/blaze/anomaly_test.clj
@@ -373,7 +373,40 @@
   (testing "anomaly"
     (given (ba/anomaly (ba/busy "msg-121702"))
       ::anom/category := ::anom/busy
-      ::anom/message := "msg-121702")))
+      ::anom/message := "msg-121702")
+
+    (testing "with an invalid category"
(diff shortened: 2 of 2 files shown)


modules/anomaly/src/blaze/anomaly.clj
```diff
@@ -151,6 +151,17 @@
...
+(defn- sanitize
+  "Replaces an invalid category of `m` with `::anom/fault` and removes an
+  invalid message.
+
+  Without that coercion, a map with an arbitrary category or a non-string
+  message would pass as an anomaly without satisfying `::anom/anomaly`."
+  [{::anom/keys [category message] :as m}]
+  (cond-> m
+    (not (categories category)) (assoc ::anom/category ::anom/fault)
+    (not (string? message)) (dissoc ::anom/message)))
```
alexanderkiel, line 163: `(not (string? message))` -> `(and (some? message) (not (string? message)))`

modules/anomaly/src/blaze/anomaly.clj
```diff
@@ -159,11 +170,9 @@
   the data of an arbitrary exception could turn the result into something that
   isn't an anomaly anymore."
   [msg data]
-  (let [{::anom/keys [category message]} data
+  (let [{::anom/keys [message]} data
         message (if (string? message) message msg)]
-    (cond-> (assoc data ::anom/category (if (categories category) category ::anom/fault))
-      (nil? message) (dissoc ::anom/message)
-      message (assoc ::anom/message message))))
+    (cond-> (sanitize data) message (assoc ::anom/message message))))
```
alexanderkiel:
```
[msg {::anom/keys [message] :as data}]
(cond-> (sanitize data)
  (and (some? msg) (nil? message)) (assoc ::anom/message message))
```

alexanderkiel (changes requested):

modules/anomaly/src/blaze/anomaly.clj
trobanga: Adopted, thanks

modules/anomaly/src/blaze/anomaly.clj
trobanga: that would let a nil message through

alexanderkiel (approved):

PR #4056, opened 2026-08-07
Generate the Grafana Dashboard from EDN Source

### LINKED ISSUE #4045: Generate the Grafana Dashboard from EDN Source
## Problem

`docs/monitoring/blaze.json` is 14,228 lines / 196 KB of hand-exported Grafana JSON. Measured over the current file:

* 102 panels and 113 queries, **all** of type `timeseries`, using only 7 units (`ops`, `s`, `bytes`, `percentunit`, `short`, `binBps`, `none`).
* The content that carries meaning — panel titles, descriptions, PromQL expressions and `legendFormat` — is about 16.7 KB, **roughly 8 % of the file**. The rest is repeated visualisation boilerplate.
* There are 13 distinct `custom` field-config blobs and 11 distinct `options` blobs across those 102 panels. The differences are drift, not intent: panel `vizConfig` versions are `8.4.4` (50 panels), `10.0.0` (6), `11.6.7` (22) and `13.0.2` (24), depending on when each panel was last touched in the UI.
* 14 panels carry a `__systemRef: hideSeriesFrom` override that was committed by accident after a legend click in the UI. `panel-100` ("Number of Index Entries 0.9 Quantile") permanently hides every series except `Observation`, twelve panels hide everything except `index`, and `panel-87` hides everything except `index-resources`.
* The file carries export noise: `metadata.uid`, `resourceVersion`, `generation: 13`, `grafana.app/updatedBy: user:efhsbuvaoyoe8f`, `saved-from-ui: Grafana v13.0.2`, a baked-in `current` value of `srv-1:8081` for the `instance` variable, and the title `"Blaze New"`.

Adding a panel means editing in the UI, re-exporting the whole file and producing a diff nobody can review.

## Goal

Make an EDN source file the thing maintainers edit. Adding a panel becomes adding one map.

`docs/monitoring/blaze.json` is removed from Git. The dashboard becomes a **release artifact**, attached to each GitHub release, and the documentation links to it there.

The output is the **Classic** dashboard JSON model rather than the `dashboard.grafana.app/v2` resource format the current file happens to be in. Classic is what the Grafana dashboards collection uses and what every Grafana from v10 through v13 accepts on import, and the dashboard uses no V2-only features (no tabs, no conditional rendering, no advanced layouts), so nothing is lost.

### New module `modules/monitoring`

Structured like any other Clojure module (`deps.edn`, `Makefile`, `tests.edn`, `src`, `test`).

* `modules/monitoring/dashboard.edn` — the source.
* `modules/monitoring/src/blaze/monitoring/dashboard.clj` — the pure transform: source data in, dashboard data out.
* `modules/monitoring/src/blaze/monitoring/dashboard/spec.clj` — `s/def` specs for the source format.
* `modules/monitoring/test/blaze/monitoring/dashboard_spec.clj` — `s/fdef` specs.
* `modules/monitoring/build.clj` — the thin I/O shell that reads the EDN, calls the transform and writes `target/blaze-dashboard.json`, invoked via `clojure -T:build gen`. The output is build output, not tracked by Git.

### Source format

```clojure
{:title "Blaze"
 :variables [...]
 :rows
 [{:title "Transaction Log"
   :panels
   [{:title "Submitted Commands"
     :description "Number of commands submitted to the transaction log."
     :unit "ops"
     :width 8
     :queries [{:expr "sum(rate(blaze_db_tx_log_submitted_commands_total{job=\"$job\",instance=\"$instance\"}[$__rate_interval]))"
                :legend "{{instance}}"}]}]}]}
```

The generator fills in the defaults: `height` 6, `width` 8, a single pinned `vizConfig` version, the shared legend/tooltip `options` and `custom` field-config block, the `${DS_PROMETHEUS}` datasource, `refId` assignment (`A`, `B`, …) and panel ids.

Grid `x`/`y` are **computed** by flowing panels left to right within a row and wrapping at 24 columns, instead of being maintained by hand. This is what turns "add two more panels" into two maps.

Where a panel genuinely needs a non-default (stacking, an axis min/max, a `decimals` setting), it gets an explicit key in the source rather than a divergent copy of the whole blob.

### Output format

The source EDN above is schema-agnostic — the output format appears in exactly one place, the emitter. It produces the Classic model:

* A flat `panels` array, each panel carrying its own `gridPos`.
* Rows as panels of `type: "row"` with `collapsed: true` and their panels nested inside. Every row in this dashboard is collapsed, so this is uniform — there is no expanded/collapsed split to handle.
* Variables in `templating.list` (`datasource`, `query` and `custom` types).
* Queries in each panel's `targets`, with `expr`, `legendFormat`, `refId` and `datasource`.
* Top-level `title`, `uid`, `annotations`, `links`, `time`, `refresh` and `schemaVersion`.

Classic is also the more stable target for generator code we maintain ourselves. V2 is still settling: the Grafana Foundation SDK carries `dashboardv2` and `dashboardv2beta1` side by side, and V1 Resource was the default for barely two minor releases (v12.2.0 to v13.0.0).

### Determinism

The generator emits JSON with sorted keys and a fixed indent, so the same source always produces the same bytes and two releases' dashboards can be diffed against each other.

### Make targets and CI

* `make -C modules/monitoring build` writes `target/blaze-dashboard.json`.
* `make -C modules/monitoring test` runs the module tests, covering the pure transform: panel count, grid flow and wrapping at 24 columns, `refId` assignment, every panel having at least one query, and units being drawn from the known set. There is no golden test against a committed file, because no generated file is committed — a generator that throws or produces a malformed dashboard fails the build instead.
* Add `monitoring` to the `module` matrix in `.github/workflows/build.yml`, sorted alphabetically between `metrics` and `module-base`.

### Metric name check

The failure mode worth guarding against is the dashboard silently charting a metric that has been renamed or removed. A new `.github/scripts/check-dashboard-metrics.sh` catches it:

* It reads the **generated dashboard JSON**, not the EDN — the Prometheus exposition format contains the full exported names (`blaze_db_node_duration_seconds_bucket`, `blaze_cache_hits_total`) literally, so extracting the identifiers from the `expr` fields with `jq` and matching them as substrings against a `/metrics` scrape needs no EDN parsing and no suffix normalisation.
* It runs in the `integration-test` job alongside the existing `.github/scripts/test-metrics.sh`, which already scrapes `http://localhost:8081/metrics` after data has been loaded.
* PromQL function names and label names are excluded via a keyword list.

The check is one-directional: it flags dashboard metrics the server does not export, not exported metrics that have no panel. The reverse direction would be noise, since plenty of metrics are intentionally not charted.

Known limitation to handle explicitly: a label-parameterised metric exports no children until first use, so a metric that the integration test never exercises is absent from the scrape and would be reported as missing even though the dashboard is correct. The script therefore carries a small allowlist of such metrics, each with a comment naming the scenario that would exercise it.

### Release artifact

The generated dashboard is attached to each GitHub release as `blaze-dashboard.json`:

* A CI job runs `make -C modules/monitoring build` and uploads the result with `actions/upload-artifact`.
* The `release` job gains a matching `actions/download-artifact` step, lists the file in `files:`, and adds that job to its `needs:`.

The `release` job currently declares `files: target/blaze-*-standalone.jar`, but performs no checkout and no artifact download, so on a clean runner that glob matches nothing — and `softprops/action-gh-release` does not fail on unmatched globs. No release has ever carried an asset, verified for v1.10.1, v1.10.0, v1.9.0, v1.8.0, v1.0.0 and v0.30.0.

The uberjar is no longer published — Blaze is distributed as a container image, and the jar exists only as an input to the image build (`build-image: uberjar` in the root `Makefile`, `COPY target/blaze-*-standalone.jar` in the `Dockerfile`). So that glob is dead configuration rather than a defect to repair: replace it with the dashboard, making `blaze-dashboard.json` the only release asset.

### Migration

The 102 existing panels are not retyped by hand. A throwaway extraction script reads `docs/monitoring/blaze.json` (before it is deleted) and emits the initial `dashboard.edn` (row title, panel title, description, unit, width, height, queries). The result is reviewed by hand, and these intentional cleanups are applied:

* Drop all 14 `__systemRef: hideSeriesFrom` overrides.
* Normalise every panel to a single `vizConfig` version.
* Drop the export metadata. The V2 resource wrapper (`metadata.uid`, `resourceVersion`, `generation`, `grafana.app/updatedBy`, `saved-from-ui`, the deprecated internal id label) has no Classic equivalent and simply goes away. Set a stable top-level `uid` and omit `id` and `version`.
* Drop the baked-in `current` values of the `job` and `instance` variables.
* Fix the title `"Blaze New"` to `"Blaze"`.

### Documentation

Delete `docs/monitoring/blaze.json`.

Rewrite the "Import the Blaze Dashboard" section of `docs/monitoring.md` to have the reader download the dashboard and upload it in Grafana's import dialog, rather than copying the contents of a file in the repository.

The link points at a specific version rather than at `latest`, built from `VITE_LATEST_RELEASE` — the same mechanism `docs/deployment.md`, `docs/database/migration.md` and the other pages already use for image tags:

```vue
<script setup lang="ts">
  const release = import.meta.env.VITE_LATEST_RELEASE;
  const dashboardUrl = `https://github.com/samply/blaze/releases/download/${release}/blaze-dashboard.json`;
</script>
```

Note that Vue does not interpolate inside HTML attributes, so a plain Markdown link containing `{{ release }}` would render the literal braces. The existing pages get away with `{{ tag }}` because it sits in `sh-vue` code fences, which are text nodes. The link therefore needs an explicit binding:

```html
<a :href="dashboardUrl">blaze-dashboard.json</a>
```

This replaces the current `[3]:` reference to `blob/main/docs/monitoring/blaze.json`.

Note for maintainers that the dashboard is generated from `modules/monitoring/dashboard.edn`.

## Acceptance Criteria

* `docs/monitoring/blaze.json` is deleted, and `make -C modules/monitoring build` produces `target/blaze-dashboard.json` from `modules/monitoring/dashboard.edn`. Running it twice produces identical bytes.
* A GitHub release carries `blaze-dashboard.json` as an asset.
* `check-dashboard-metrics.sh` runs in the `integration-test` job and passes. Renaming a metric in the dashboard without renaming it in the code makes it fail.
* The generated dashboard imports into Grafana 13 and all 102 panels render with data, matching the current dashboard row by row. Being Classic, it also imports into Grafana v10 and later.
* `docs/monitoring.md` links to the release asset of a specific version, resolved from `VITE_LATEST_RELEASE`, and no longer references a file in the repository. The rendered page shows a working versioned URL, not literal `{{ }}` braces.
* `make fmt`, `make lint`, `make -C modules/monitoring test` and `make -C modules/monitoring test-coverage` (≥ 95 % forms) pass.
* `monitoring` is present in the CI module matrix.

## Out of Scope

* Round-tripping UI edits back into EDN. The EDN is the source of truth; UI edits are for experimenting and get ported by hand.
* Generating panels from the `defcounter` / `defhistogram` definitions. Only 15 of the 75 metric names the dashboard uses come from those macros. The other 60 are built as runtime string arguments inside collectors (41 `blaze_rocksdb_*`, 6 `blaze_cache_*` / `blaze_page_store_*`, 2 `thread_pool_executor_*`), come from operation modules (3 `fhir_*`), or are third-party (8 `jvm_*` / `process_*` from `simpleclient_hotspot`). A definition also does not determine a panel: `blaze_db_resource_indexer_duration_seconds` alone drives five, differing in quantile, aggregation, unit and row. The metric-name check above covers the part of this idea that pays off.
* The two new panels themselves. They get added on top of this, in a follow-up issue.

## Note on Tooling

A hand-written emitter is used rather than the Grafana Foundation SDK. The SDK's stable Java artifacts on Maven Central do cover the Classic model and could be called from Clojure via `deps.edn`, so this is a judgment call rather than a blocked door. Against it: the SDK is public preview at v0.0.x with no SLA, its published versions carry timestamp suffixes (`11.6.0-1769699379`) that make pinning awkward, and its Java builder chains read poorly from Clojure. With one panel type and 7 units in play, the mapping we own is small enough that a data-oriented Clojure emitter is the better fit for this repository.

alexanderkiel:
Closes: #4045


docs/monitoring.md
```diff
@@ -1,3 +1,8 @@
+<script setup lang="ts">
+  const release = import.meta.env.VITE_LATEST_RELEASE;
```
knoppiks, line 2: const appears unused.

modules/monitoring/dashboard.edn
```diff
@@ -0,0 +1,778 @@
+;; Source of the Blaze Grafana dashboard.
+;;
+;; Generated into target/blaze-dashboard.json by `make -C modules/monitoring build`.
+{:title "Blaze"
+ :uid "Q-h9isMWk"
```
knoppiks: Do we really have to use the uid from the copied json dashboard?
`blaze-server.org--backend` or the like would be nice.

knoppiks (changes requested):
Did you check https://github.com/grafana/dashboard-linter - could be useful to check the resulting dashboard against best-practices. - if they are not already checked for in the tests.

docs/monitoring.md
alexanderkiel: Is used in `dashboardUrl`

modules/monitoring/dashboard.edn
alexanderkiel: I removed the UID.

knoppiks (approved):

PR #4088, opened 2026-08-13
Fix Error After Frontend Session Expiry

### LINKED ISSUE #4082: Fix Error After Frontend Session Expiry
Currently, an expired or rejected session can still end in a 500 Internal Error page instead of the sign-in page. Reloading the browser then does redirect to sign-in, so the user has to guess that a reload is what is needed.

An expired session should always lead to the sign-in page, whether the request is a page load or a navigation inside the running app.

Why: a 500 tells the user the server is broken when in fact they are merely signed out, and the recovery step is undiscoverable.

knoppiks:
Closes: #4082

#4163 moved every load server side, which removed the reported symptom: the universal loads that fetched a `+server.ts` proxy with a plain browser `fetch`, silently followed the 307 to the sign-in page and choked on its HTML in `res.json()`. Every request now either is a document load, SvelteKit's own `__data.json` request or a `use:enhance`d submission — all of which SvelteKit turns into a navigation when answered with a redirect.

What that does not reach is the case `handleAuthorization` cannot see: the Auth.js session is valid, but the backend rejects the access token — revoked at the IdP, clock skew, a realm mismatch. Today that 401 reaches the load and becomes an error page.

So `handleFetch` redirects on a backend 401, reusing the `signInUrl` from #4144. It has to happen in the hook rather than in the load, because `event.url` — the page the user is on, and so the return-to target — is not something a load can read. A 403 is passed through: there the token was accepted and signing in again changes nothing.

One consequence, documented rather than fixed: a load that streams must await at least one backend request first. A redirect thrown after `load` has returned cannot become a navigation; SvelteKit serializes it into the streamed chunk as a 500. Both streamed loads already await `fetchSearchMetadata`, and `/metadata` is behind the same `wrap-auth-guard` as everything else, so a rejected token is always discovered while a redirect still works. The StructureDefinition loads inside `transformBundle` remain the narrow exception.

The e2e test covers the `handleAuthorization` half — it clears the session cookie, so no session exists and no backend request is made. What it adds over the unit tests is the envelope: it is the first test to exercise the client router turning a redirect found in a `__data.json` response into a navigation, which #4163 verified only by reading `@sveltejs/kit`. Reaching `handleFetch` from Playwright would mean revoking the Keycloak session behind the app's back; `src/hooks.server.test.ts` covers it instead.

### DIFF
--- a/docs/implementation/frontend.md
+++ b/docs/implementation/frontend.md
@@ -34,6 +34,17 @@ request was made. SvelteKit puts it into the right envelope by itself:
 | `__data.json` data request     | `{ type: 'redirect' }`, which the client router turns into a `goto`   |
 | `use:enhance`d form submission | a redirect `ActionResult`, which `applyAction` turns into a navigation |
 
+### Rejected access tokens
+
+`handleAuthorization` only knows what the Auth.js session claims, which is its `expiresAt`. The backend can still reject 
+the access token that session carries — revoked at the identity provider, clock skew, a realm mismatch — and
+answers 401. `handleFetch` redirects to the sign-in page in that case, for the same reason `handleAuthorization` can: 
+the request it was serving is one SvelteKit turns into a navigation.
+
+The recovery lives in the hook rather than in the load that made the request, because `event.url` — the page the 
+user is on, and so the return-to target — is not something a load can read. A 403 is passed through instead:
+there the token was accepted, and the user simply may not do this, which signing in again does not change.
+
 ### Backend URLs
 
 Backend request URLs are built by prefixing `base` from `$app/paths`, for example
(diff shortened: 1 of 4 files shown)


docs/implementation/frontend.md
```diff
@@ -13,5 +13,69 @@ The frontend uses the [Auth.js][1] [@auth/sveltekit][2] library for for authenti
 * the access token will be refreshed via the refresh token if possible
 * the session will expire at the same time as the last successful refreshed access token will expire 
 
+## Session Expiry
+
+When the session expires, the user has to end up on the sign-in page and be returned to the page they came from
```
alexanderkiel: This sounds like a requirement. I would reword it so that it describes the current state of the implementation.

modules/frontend/src/lib/sign-in.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  return `${signInPath}?redirect=${encodeURIComponent(target.pathname + target.search)}`;
+}
+
+/**
+ * A 401 response is how the server signals that the session has expired
+ * (RFC 6750 section 3). See `canRedirectToSignIn` in `hooks.server.ts` for
+ * the other half of this contract.
+ */
+export function sessionExpired(res: Pick<Response, 'status'>): boolean {
```
alexanderkiel: Should be named `isSessionExpired`

modules/frontend/src/hooks.server.ts
```diff
@@ -19,6 +19,17 @@ function signInUrl(target: URL): string {
...
 }
 
+/**
+ * A 401 on a Bearer-protected resource means exactly that (RFC 6750 section
+ * 3), whatever challenge accompanies it — so the status alone is enough, and
+ * a proxy in front of Blaze that omits `WWW-Authenticate` is covered too. A
+ * 403 is deliberately not included: there the token was accepted, and the user
+ * simply may not do this, which signing in again does not change.
+ */
+function sessionExpired(res: Response): boolean {
```
alexanderkiel: IHMO this function isn't necessary. The doctoring is overly complex and it has also nothing to do with the session. Instead, just test for 401 where this function is called. The comment there is completely fine.

alexanderkiel (changes requested):

docs/implementation/frontend.md
alexanderkiel: outdated

modules/frontend/src/lib/sign-in.ts
alexanderkiel: outdated

alexanderkiel (approved):

Issue #4101, opened 2026-08-18
Pipeline Resource Indexing Across Transactions


alexanderkiel:
The database node indexes one transaction at a time. `index-tx-data!` dispatches the resource indexing of a transaction to the resource indexer executor, indexes the transaction itself, stores the transaction entries and then blocks in `wait-for-resources` until the resources of that very transaction are indexed, before it commits and moves on to the next transaction. Nothing overlaps across transactions, although `poll!` already delivers up to 50 transactions at once.

## Measurement

Measured on LEA79 at 5300 tx/s with the transaction load test, which submits two resources per transaction. Of the 189 µs per transaction:

| | per tx | |
|---|---|---|
| `index-resources` (wall, brackets the next two) | ~181 µs | |
| ├ `index-transactions` | ~28 µs | overlapped |
| ├ `store-tx-entries` | ~14 µs | overlapped |
| └ blocked in `ac/join` | **~139 µs** | 74 % of the loop |
| `store-tx-success-entries` | ~7 µs | |
| `poll-tx-log` | ~0.9 µs | |

The `index-resources` histogram brackets `index-transactions` and `store-tx-entries`, so the values are not additive. They are derived from the shares of the measured sum and should be re-read from `rate(_sum)/rate(_count)` before they are used as a baseline.

Because a transaction of the load test contains only two resources, the resource indexing fans out only two tasks, so at most two of the four threads of the resource indexer executor are ever active. Raising `DB_RESOURCE_INDEXER_THREADS` cannot recover the idle time — the serialization itself has to go.

Goal: peak throughput under concurrency, at the upper end of the load test sweep. Single transaction latency at concurrency 1 is unaffected either way.

## Second problem: unbounded fan-out per transaction

Independent of the serialization, the resource indexing of a single transaction is unbounded in two ways. `index-resources` fans out one executor task per resource, and on a node without local payload it first fetches every resource of the transaction in one `rs/multi-get`. The Cassandra resource store doesn&#39;t batch that call, it issues one query per key:

https://github.com/samply/blaze/blob/main/modules/db-resource-store-cassandra/src/blaze/db/resource_store/cassandra.clj#L80-L81

So a transaction with 100.000 resources issues 100.000 concurrent Cassandra queries and holds every decoded resource in memory before a single index entry is written. Both the query concurrency and the resource memory scale with the transaction size and nothing bounds them.

[...]

PR #4110, opened 2026-08-20
Cache Type and System Stats in the Indexer

### LINKED ISSUE #4109: Cache Type and System Stats in the Indexer
The TypeStats and SystemStats indexes are maintained with a read-modify-write per transaction. For every resource type touched, `type-stat-entry` seeks the previous counter value out of RocksDB, adds the increments of the transaction and emits a new entry; `system-stats` does the same once for the system-wide totals:

https://github.com/samply/blaze/blob/main/modules/db/src/blaze/db/node/tx_indexer/verify.clj#L295-L311

Both seeks go through `i/seek-value`, which opens an iterator on the batch database snapshot, seeks and closes it again. For a small transaction touching two resource types that is three seeks — one per type plus the system totals — on a thread that is fully saturated.

The values being seeked are values the indexer itself wrote. `verify.clj` is the only writer of both column families; every other use is a read at an arbitrary `t` from `batch-db` (`type-count`, `system-count`, `total-num-of-type-changes`, `total-num-of-system-changes`). And the indexer applies transactions strictly in order on a single thread, so the entry it seeks at the `t` of `db-before` is always exactly the one it wrote for the previous transaction. The seek re-reads state the loop already knows.

## Measurement

Wall clock profile of `index-batch!` on an enterprise AMD EPYC™ server with fast enterprise SSDs (LEA79) with the transaction load test, 128 VUs, 5 minutes, at about 15,000 tx/s. `index-batch!` covers 98 % of the indexer thread, which puts the budget at 65.3 µs per transaction. Each transaction touches two resource types, so the seek counts below are two types plus the system totals.

| | share | per tx |
|---|---|---|
| `post-process-res` | 19.8 % | 12.9 µs |
| ├ TypeStats `seek` (2 types) | 11.9 % | 7.8 µs |
| ├ SystemStats `seek` | 5.1 % | 3.3 µs |
| └ `into` / `transduce` / `merge-with +` | 2.8 % | 1.8 µs |

On top of that both column families pay for a fresh iterator per transaction, because the iterator pool lives on the snapshot and every transaction gets a new batch database: 1.1 % for the native `RocksDB.iterator` of the TypeStats iterator and 1.0 % for the SystemStats one, plus their share of the 3.4 % spent disposing iterators in `BatchDb.close`.

## Proposal

Keep the current counters in memory in the node:

* a map of `tid` to `{:total, :num-changes}` for the types, and one `{:total, :num-changes}` for the system,
* seeded once at startup from the indexes at the current `t` — one seek per type that has an entry plus one for the system,
* read by `post-process-res` instead of seeking, and
* updated after `store-tx-entries!` succeeded, so that a rejected transaction leaves the counters untouched.

The index entries themselves keep being written exactly as they are today, because `batch-db` reads them at arbitrary `t` for history and search totals. Only the read of the head value moves into memory.

## Expected Effect

About 17 % of the indexer loop for the seeks, plus most of the iterator creation and disposal of the two column families and the `merge-with +` allocation, which the in-memory counters can do on longs. With the loop saturated, a saving on it converts into throughput roughly one to one.

Part of #4103.

alexanderkiel:
Closes: #4109


alexanderkiel:
## Achieved Speedup

Transaction load test, measured against the baseline of #4101 (`c59dfaa`), 5 minutes per level.

### LEA79 (128 cores, local NVMe, syncs from write cache)

| VUs | Req/s before | Req/s after | Δ | med before | med after | q99 before | q99 after |
|----:|-------------:|------------:|----:|-----------:|----------:|-----------:|----------:|
| 1 | 1,275 | 1,179 | −8 % | 0.5 ms | 0.6 ms | 0.8 ms | 0.8 ms |
| 2 | 2,610 | 2,505 | −4 % | 0.5 ms | 0.5 ms | 0.7 ms | 0.7 ms |
| 4 | 4,771 | 4,455 | −7 % | 0.6 ms | 0.6 ms | 0.9 ms | 0.9 ms |
| 8 | 7,192 | 7,536 | +5 % | 0.8 ms | 0.8 ms | 1.1 ms | 1.1 ms |
| 16 | 11,117 | 12,013 | +8 % | 1.1 ms | 1.1 ms | 1.6 ms | 1.5 ms |
| 32 | 13,309 | 15,757 | **+18 %** | 2.1 ms | 1.8 ms | 2.9 ms | 2.3 ms |
| 64 | 14,739 | 17,833 | **+21 %** | 3.9 ms | 3.2 ms | 5.9 ms | 6.1 ms |
| 128 | 15,392 | 17,968 | **+17 %** | 7.8 ms | 6.6 ms | 11.8 ms | 10.9 ms |

### LEA47 (16 cores, vSAN, 484 fsyncs/s)

| VUs | Req/s before | Req/s after | Δ | med before | med after | q99 before | q99 after |
|----:|-------------:|------------:|----:|-----------:|----------:|-----------:|----------:|
| 1 | 179 | 178 | −1 % | 5.2 ms | 5.2 ms | 6.8 ms | 6.9 ms |
| 2 | 369 | 368 | ±0 % | 5.0 ms | 5.0 ms | 7.1 ms | 7.3 ms |
| 4 | 464 | 479 | +3 % | 7.8 ms | 7.5 ms | 11.6 ms | 11.2 ms |
| 8 | 807 | 848 | +5 % | 9.8 ms | 9.4 ms | 13.5 ms | 12.8 ms |
| 16 | 1,464 | 1,499 | +2 % | 10.3 ms | 10.1 ms | 18.3 ms | 17.1 ms |
| 32 | 2,198 | 2,582 | **+18 %** | 13.7 ms | 11.4 ms | 28.9 ms | 25.4 ms |
| 64 | 2,429 | 3,137 | **+29 %** | 24.9 ms | 19.1 ms | 55.1 ms | 46.5 ms |
| 128 | 2,637 | 3,313 | **+26 %** | 47.1 ms | 37.3 ms | 79.2 ms | 75.4 ms |

### A5N46 (consumer NVMe, 178 fsyncs/s)

Re-run as well, with no measurable difference, so its CSV is unchanged. A5N46 doesn't spend its time in the indexer loop — it spends it storing the transaction data. Its drive acknowledges only 178 syncs per second, so every transaction waits for the durable write of its log entry and its resources, and that wait sets both the processing time and the throughput. The indexer thread has slack the whole time, so taking the two seeks out of it converts into nothing: the loop was never what the transactions were queuing behind. That is also what its numbers show — the median pinned at 29 ms and the q99 at 41 ms across the whole sweep, with throughput growing only through larger group commits.

### Reading of the numbers

The gain appears exactly where #4109 predicted it: on the systems whose indexer loop is the bottleneck, and only once it runs saturated. Below 16 VUs neither LEA47 nor LEA79 saturates it — the rate is set by the disk and the client concurrency there, and the differences at those levels are run-to-run noise. From 32 VUs on, throughput rises by 17–29 %, which matches the ~17 % of `index-batch!` the two seeks and their iterators accounted for, plus the `merge-with +` allocation the in-memory counters now do on longs.

LEA79 moves its plateau up rather than just reaching it earlier: it flattens at about 18,000 transactions/s from 64 VUs on instead of 15,000. On LEA47 the effect is larger in relative terms because its indexer thread is the scarcer resource of the two — the saturation point moves from 32 to 64 VUs, so it now converts one more doubling of clients into throughput before the clients only queue.

Latencies improve along with it at those levels — the median at 128 VUs drops from 7.8 ms to 6.6 ms on LEA79 and from 47 ms to 37 ms on LEA47 — since the queue ahead of the indexer is what those clients wait in.

The descriptions in `docs/performance/load-testing.md` and the LEA79 cross-reference in `docs/performance/disk-perf.md` are updated to the new numbers.

modules/db/test/blaze/db/node_test.clj
```diff
@@ -531,6 +531,24 @@
...
               ::x := nil)))))))
 
+(deftest stats-test
+  (testing "a rejected transaction leaves the type and system stats untouched,
+            so that the transaction after it continues from the ones before it"
+    (with-system [{:blaze.db/keys [node]} config]
+      @(d/transact node [[:put {:fhir/type :fhir/Patient :id "0"}]])
+
+      (given-failed-future
+       (d/transact node [[:put {:fhir/type :fhir/Patient :id "0"} [:if-match 5]]])
```
knoppiks, line 541: Why 5?

knoppiks (approved):

modules/db/test/blaze/db/node_test.clj
alexanderkiel: Why not? It's just a `t` which has to be different to 1.

PR #4149, opened 2026-08-28
Unify Backend URL Generation

### LINKED ISSUE #4148: Unify Backend URL Generation
Currently, the frontend builds the URLs of its backend requests in three different ways: through `resolve()`, which is SvelteKit's helper for building links and navigation targets from route IDs; through a hardcoded `/fhir` prefix; and by prefixing the configured base path by hand. The hardcoded ones break as soon as the application is mounted somewhere other than `/fhir`, and the `resolve()` ones read as if a page of this app were being requested when in fact the backend is.

There should be one way to build a backend request URL, distinct from the way navigation targets are built, so that reading a call site tells you which of the two is happening.

Why: the two kinds of URL only look interchangeable. They are built from different inputs and they are not always equal, so conflating them hides the cases where they differ.

knoppiks:
Closes: #4148

### DIFF
--- a/modules/frontend/src/lib/backend.ts
+++ b/modules/frontend/src/lib/backend.ts
@@ -1,0 +1,31 @@
+import { base } from '$app/paths';
+
+/**
+ * Builds the URL of a backend request from `path` and an optional `query`.
+ *
+ * This is the only way backend request URLs are built. It is deliberately
+ * distinct from `resolve` of `$app/paths`, which builds navigation targets
+ * from route IDs: the two kinds of URL are built from different inputs and
+ * are not always equal, so a call site should say which of the two it means.
+ *
+ * The result is root-relative on purpose. `handleFetch` in `hooks.server.ts`
+ * swaps this app's origin for `BACKEND_BASE_URL` and keeps the path, so the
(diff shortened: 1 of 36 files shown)


modules/frontend/src/lib/backend.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * The result is root-relative on purpose. `handleFetch` in `hooks.server.ts`
+ * swaps this app's origin for `BACKEND_BASE_URL` and keeps the path, so the
+ * base path is part of the backend URL as well, and an absolute URL would have
+ * to duplicate that knowledge.
+ *
+ * `base` is deprecated in favor of `resolve`, but `resolve` cannot serve this
+ * case, which is why this stays the single place that reads it:
+ * - Most backend paths are not routes of this app at all — there is none for
+ *   `/__admin/Task/{id}/$pause` or `/GraphDefinition`. That the generated
```
alexanderkiel: `/GraphDefinition` is a valid frontend path

alexanderkiel (changes requested):

alexanderkiel (approved):

PR #4166, opened 2026-08-31
Use Server Side Loading Only

### LINKED ISSUE #4163: Use Server-Side Loading Only
Currently the frontend loads its data in two ways. Most load functions run in the browser and reach the backend through endpoints the frontend serves only in order to forward the request; the rest run on the server and call the backend directly. Those forwarding endpoints exist so that the browser can reach the backend through the frontend server — which is exactly what the frontend server already does by itself for a load that runs there.

All loading should happen server side, and there should be no endpoints whose only purpose is forwarding.

Why:

* The two paths are not equivalent. A request the browser makes on its own is not a navigation, so when the server answers it with a redirect — asking the user to sign in again, for example — the browser follows it silently and the load is handed a page of HTML where it expected data. Loading server side removes that whole class of failure instead of trying to detect it. This is the failure mode behind #4082, though what remains of that issue is a separate case.
* A forwarding endpoint has to mirror the backend's URL space to work at all, so every backend path ends up written down twice and the two copies have to be kept in step.
* It costs a round trip on every load: the browser asks the frontend, which asks the backend.

knoppiks:
Closes: #4163


alexanderkiel (changes requested):
We need the status in the error body in metadata.ts. So we need an appError there.

alexanderkiel (approved):

PR #4184, opened 2026-09-03
Improve Parameter Coercion in Terminology Service

### LINKED ISSUE #4178: Improve Parameter Coercion in Terminology Service
The in parameters of the terminology service operations are coerced inconsistently. Boolean parameters aren't type checked at all, and the integer coercion open codes the value extraction in a way that throws on a wrong type instead of returning an anomaly.

## Boolean In Parameters Aren't Coerced

The boolean in parameters of `$expand` and `$validate-code` are declared as bare `{:action :copy}` in `blaze.terminology-service.local`:

* `vs-expand-param-specs`: `includeDesignations`, `includeDefinition`, `activeOnly`, `excludeNested`
* `vs-validate-code-param-specs`: `inferSystem`, `lenient-display-validation`, `activeOnly`

Without `:coerce`, `fu/coerce-params` copies the value of the primitive unchecked. So a `Parameters` resource carrying `valueString` instead of `valueBoolean` is accepted and the `String` flows on to consumers that expect a `Boolean`:

* `value_set/validate_code.clj:222` — `(if (or active-only (false? inactive)) ...)`. `"false"` is truthy, so `activeOnly` with a string value behaves like `activeOnly=true`.
* `value_set/validate_code/issue.clj:99` — `(if lenient-display-validation ...)`, the same inversion.
* `value_set/expand.clj:104-114` — `(type/boolean active-only)` on a `String` returns an anomaly, which then ends up in `ValueSet.expansion.parameter.value`.

Those seven params should be coerced, so that a wrong type is rejected with a `400` the way it already is for `offset` and `count`.

## Integer Coercion Throws on a Wrong Type

`coerce-nat-long` in `blaze.terminology-service.local` extracts the value itself:

```clojure
(defn- coerce-nat-long [value]
  (if-some [value (:value value)]
    (if-not (neg? value)
      value
      (ba/incorrect "Has to be a non-negative integer."))
    (ba/incorrect "Missing value.")))
```

For `count` or `offset` given as `valueString`, `(:value value)` is a `String` and `neg?` throws a `ClassCastException` instead of returning an anomaly. The value extraction belongs in a shared coercer in `blaze.fhir.util` that checks the type first, leaving `coerce-nat-long` with just the non-negative check.

## Naming of the Query Param Coercers

`fu/coerce-boolean` and `fu/coerce-integer` don't coerce a FHIR value — they take a parameter name plus the string value of a query param and build a FHIR value from it. Now that coercers for the values of a `Parameters` resource are needed as well, the query param ones should be renamed to `fu/coerce-query-boolean` and `fu/coerce-query-integer`, freeing the plain names for the `Parameters` side. The call sites are `value_set/expand.clj` and `value_set/validate_code.clj`.

A coercer for string valued params (`url`, `code`, `version`, `display`, `filter`, `valueSetVersion`, …) could follow the same way if the same type check is wanted there.

## Tests

* no test covers a non-integer `count`/`offset`
* no test covers a wrong typed value for any of the boolean params

alexanderkiel:
Closes: #4178

### DIFF
--- a/modules/rest-util/src/blaze/handler/fhir/util_spec.clj
+++ b/modules/rest-util/src/blaze/handler/fhir/util_spec.clj
@@ -3,6 +3,7 @@
    [blaze.async.comp :as ac]
    [blaze.db.spec]
    [blaze.fhir.spec]
+   [blaze.fhir.spec.type :as type]
    [blaze.handler.fhir.util :as fhir-util]
    [blaze.handler.fhir.util.spec]
    [blaze.http.spec]
@@ -126,3 +127,15 @@
                                       :blaze.preference/return])
                :entries (s/coll-of :fhir.Bundle/entry))
   :ret (s/or :response-entry :fhir.Bundle/entry :anomaly ::anom/anomaly))
+
+(s/fdef fhir-util/coerce-query-boolean
+  :args (s/cat :name string? :value string?)
+  :ret (s/or :value type/boolean? :anomaly ::anom/anomaly))
+
+(s/fdef fhir-util/coerce-query-integer
+  :args (s/cat :name string? :value string?)
+  :ret (s/or :value type/integer? :anomaly ::anom/anomaly))
+
+(s/fdef fhir-util/validate-query-params
+  :args (s/cat :specs map? :parameters :ring.request/query-params)
+  :ret (s/or :parameters :fhir/Parameters :anomaly ::anom/anomaly))
--- a/modules/fhir-structure/src/blaze/fhir/util.clj
+++ b/modules/fhir-structure/src/blaze/fhir/util.clj
@@ -150,6 +150,53 @@
     (str (subs s 0 (dec (count s))) "ies")
     (str s "s")))
 
+(defn coerce-integer
+  "Returns the int value of the integer `x`.
+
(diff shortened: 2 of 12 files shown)


modules/rest-util/src/blaze/handler/fhir/util_spec.clj
```diff
@@ -126,3 +127,15 @@
...
+  :args (s/cat :name string? :value string?)
+  :ret (s/or :value type/boolean? :anomaly ::anom/anomaly))
+
+(s/fdef fhir-util/coerce-query-integer
+  :args (s/cat :name string? :value string?)
+  :ret (s/or :value type/integer? :anomaly ::anom/anomaly))
+
+(s/fdef fhir-util/validate-query-params
+  :args (s/cat :specs map? :parameters :ring.request/query-params)
+  :ret map?)
```
knoppiks:
```suggestion
  :ret (s/or :parameters :fhir/Parameters :anomaly ::anom/anomaly))
```

knoppiks: `fu/coerce-params` has the same spec.

knoppiks (changes requested):

knoppiks (approved):

PR #4186, opened 2026-09-03
Extract Fetch and Error Handling From Loads

### LINKED ISSUE #4150: Extract Fetch and Error Handling From Loads
Currently, every load function in the frontend repeats the same four steps by hand: fetch with an explicit Accept header, check whether the response is ok, build an error body and abort, then parse the JSON. The error bodies are built inline, so the same "not found / deleted, look into the history / generic failure" wording is written out again in each load that reads a single resource, with small unintended differences between copies.

Loading should be expressed once, with each load supplying only what is specific to it: the URL and the error message.

Why: the repetition is where the inconsistencies come from, and it is what makes any change to how loads report failures a change to twenty files.

knoppiks:
Closes: #4150

### DIFF
--- a/modules/frontend/src/lib/fetch.ts
+++ b/modules/frontend/src/lib/fetch.ts
@@ -1,0 +1,152 @@
+import { error, type NumericRange } from '@sveltejs/kit';
+
+/**
+ * The `App.Error` a load aborts with when its request fails, or a way to get
+ * one.
+ *
+ * Most loads know their message up front, so a plain string or an `App.Error`
+ * is enough. A function is for the loads whose body depends on the response,
+ * which is what `loadError` builds.
+ */
+export type ErrorSource =
+  string | App.Error | ((res: Response) => App.Error | string | Promise<App.Error | string>);
+
+/**
+ * The message a load reports for one status.
+ *
+ * A function is for a message only in the response, like the
+ * diagnostics of an `OperationOutcome`.
+ */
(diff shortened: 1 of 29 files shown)


modules/frontend/src/lib/metadata.ts
```diff
@@ -1,43 +1,32 @@
...
   });
 
-  if (!res.ok) {
-    const status = res.status as NumericRange<400, 599>;
-    error(status, appError(status, `error while loading the ${type} StructureDefinition`));
-  }
-
-  const bundle = (await res.json()) as Bundle;
-
+  // The status is carried in the body as well, like `fetchFhir` does it,
```
alexanderkiel: comment should be removed

modules/frontend/src/lib/fetch.ts
```diff
@@ -0,0 +1,153 @@
...
+    error(status, { status, ...(await appErrorOf(options.error, res)) });
+  }
+
+  return (await res.json()) as T;
+}
+
+/**
+ * Reads the FHIR resource at `url`.
+ *
+ * This is how a load reads from the FHIR API — every resource endpoint,
```
alexanderkiel: remove that overspecific examples

alexanderkiel: and document the thrown execptions

alexanderkiel (changes requested):

